PR #43, opened 2024-07-15
[20] Connecting Rails app to Elixir Dev Database

### LINKED ISSUE #20: Connecting Rails app to Elixir Dev Database
As a developer, I would like to persist data on the new platform in the same database that the portal uses.

_Note_: this will make migrations more complicated. The migration approach will be defined in #21.

Acceptance criteria:

- [x] When this ticket is complete, the Rails app's development environment will be connected to the same Postgresql database that the Elixir app uses.

cpreisinger:
#20


Gemfile
```diff
@@ -8,8 +8,8 @@ gem "rails", "~> 7.1.3", ">= 7.1.3.4"
 # The original asset pipeline for Rails [https://github.com/rails/sprockets-rails]
 gem "sprockets-rails"
 
-# Use sqlite3 as the database for Active Record
-gem "sqlite3", "~> 1.4"
+# Use postgresql as the database for Active Record
+gem "pg"
```
stepchud, line 12: can you add your commit `Gemfile.lock`? I have a few changes locally but I'm not sure which ones are needed here

db/schema.rb
stepchud, line 1:
```
PG::UndefinedTable: ERROR:  relation "challenge_owners_id_seq" does not exist (ActiveRecord::StatementInvalid)
```

I was investigating this local schema error when running `bin/rails db:prepare` and found this [Rails guide article](https://edgeguides.rubyonrails.org/active_record_migrations.html#types-of-schema-dumps) which recommends to use the `:sql` dumper instead of `:ruby` schema_format. this will let us dump & load postgresql-specific features like the `nextva()` function which is used for bigint ID sequences. I'll push this change to your branch since it's easier than describing what to do.

config/application.rb
```diff
@@ -23,5 +23,8 @@ class Application < Rails::Application
     #
     # config.time_zone = "Central Time (US & Canada)"
     # config.eager_load_paths << Rails.root.join("extras")
+
+    # Use the Postgresql-specific syntax for DB dumps
+    config.active_record.schema_format = :sql
```
stepchud, line 28: change the default format for schema dumps to support postgresql-specific features

stepchud:
```
PG::UndefinedTable: ERROR:  relation "challenge_owners_id_seq" does not exist (ActiveRecord::StatementInvalid)
```

I was investigating this local schema error when running `bin/rails db:prepare` and found this [Rails guide article](https://edgeguides.rubyonrails.org/active_record_migrations.html#types-of-schema-dumps) which recommends to use the `:sql` dumper instead of `:ruby` schema_format. this will let us dump & load postgresql-specific features like the `nextva()` function which is used for bigint ID sequences.

stepchud (approved):
looks good to me! 🎸

PR #44, opened 2024-07-16
[3] USWDS Design System 

### LINKED ISSUE #3: Install and configure the latest USWDS version
As a challenge platform developer, I want to use the latest design system components for the new project.

Acceptance criteria:

- [x] The Rails app has the latest USWDS installed (currently: 3.8.1). 
- [x] The design system is available for use in the app views and templates.

https://designsystem.digital.gov/how-to-use-uswds/

stonefilipczak:
Adds USWDS design system. https://github.com/GSA/Challenge_platform/issues/3

Sets up a /sandbox route for demonstrating some components.

Instructions followed: https://designsystem.digital.gov/documentation/getting-started/developers/phase-two-compile/#step-1-install-uswds-compile-2


.gitignore
```diff
@@ -34,3 +34,10 @@
 
 # Ignore master key for decrypting credentials and more.
 /config/master.key
+
+node_modules
```
stepchud: do we need this with the `/node_modules` below?

package-lock.json
stepchud, line 1: do we need both package-lock.json and yarn.lock? I think we can just use yarn for package management and not mess with npm

gulpfile.js
```diff
@@ -0,0 +1,30 @@
...
+uswds.settings.version = 3;
+
+/**
+* Path settings
+* Set as many as you need
+*/
+uswds.paths.dist.css = './app/assets/stylesheets';
+uswds.paths.dist.theme = './app/assets/uswds';
+uswds.paths.src.sass = './node_modules/@uswds/uswds/packages'
+uswds.paths.dist.img = '../app/img'
```
stepchud: is this path correct using the parent directory?

app/javascript/application.js
```diff
@@ -0,0 +1,3 @@
+console.log("tweet")
```
stepchud: 🧐

config/routes.rb
```diff
@@ -7,4 +7,5 @@
 
   # Defines the root path route ("/")
   # root "posts#index"
+  resources :sandbox
```
stepchud: it might be nice to keep a sandbox route for development-only purposes, like this
```suggestion
  if Rails.env.development?
    get "/sandbox", to: "sandbox#index"
  end
```

stepchud (approved):
works for me, I'll merge it once the build passes

PR #46, opened 2024-07-18
[38] Create Rails User model

### LINKED ISSUE #38: Create Rails User model
As a Rails developer, I want a User model to define behavior related to users in the system.

The User model should mirror the behavior of the `ChallengeGov.Accounts.User` schema in the Phoenix app.
The `users` schema already exists in the DB and should not require any migration for this ticket.

bilalhankins:
- Replicated user model from challenge_gov Elixir schema using the [schema_to_scaffold gem](https://github.com/frenesim/schema_to_scaffold)


db/migrate/20240718185343_create_users.rb
stepchud, line 1: remove

app/controllers/users_controller.rb
stepchud, line 1: remove

app/views/users/_form.html.erb
stepchud, line 1: remove

app/views/users/_user.html.erb
stepchud, line 1: remove

app/views/users/_user.json.jbuilder
stepchud, line 1: remove

app/views/users/edit.html.erb
stepchud, line 1: let's remove everything under `app/views/users/`

config/routes.rb
```diff
@@ -1,4 +1,5 @@
 Rails.application.routes.draw do
+  resources :users
```
stepchud: remove

spec/requests/users_spec.rb
stepchud, line 1: remove

spec/routing/users_routing_spec.rb
stepchud, line 1: remove

spec/views/users/edit.html.erb_spec.rb
stepchud, line 1: remove everything under `spec/views/users/`

stepchud (changes requested):
Now that I've seen this, it doesn't look very useful. I'm not even sure we need the view helpers TBH. I think all we really want are the app/models and spec/models classes but those can be created pretty easily without the `schema_to_scaffold` gem. instead let's try using https://github.com/ctran/annotate_models to add code comments to the model class based on the current schema.

spec/helpers/users_helper_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+# describe UsersHelper do
+#   describe "string concat" do
+#     it "concats two strings with spaces" do
+#       expect(helper.concat_strings("this","that")).to eq("this that")
+#     end
+#   end
+# end
+RSpec.describe UsersHelper, type: :helper do
+  pending "add some examples to (or delete) #{__FILE__}"
+end
```
stepchud: since the helper file is empty, you can either remove both the helper file and this spec, or just delete the pending spec
```suggestion
```

spec/models/user_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+#  status                     :string(255)
+#  active_session             :boolean          default(FALSE)
+#  renewal_request            :string(255)
+#  jwt_token                  :text
+#  recertification_expired_at :datetime
+#
+require 'rails_helper'
+
+RSpec.describe User, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
stepchud: let's add some basic specs to start out with, like maybe test it validates the presence of email and active_session is false by default

app/models/user.rb
```diff
@@ -0,0 +1,82 @@
...
+  attribute :avatar_key, :uuid
+  attribute :avatar_extension, :string
+
+  attribute :terms_of_use, :datetime
+  attribute :privacy_guidelines, :datetime
+  attribute :agency_id, :integer
+
+  attribute :last_active, :datetime
+  attribute :recertification_expired_at, :datetime
+  attribute :active_session, :boolean
```
stepchud: this is default: false in the DB

app/models/user.rb
```diff
@@ -0,0 +1,82 @@
...
+  has_many :challenge_manager_challenges, through: :challenge_managers, source: :challenge
+  has_many :members
+  has_many :supporting_documents, class_name: 'Document'
+  has_many :submissions, foreign_key: :submitter_id
+  has_many :managed_submissions, class_name: 'Submission', foreign_key: :manager_id
+  has_many :submission_documents, class_name: 'Submissions::Document'
+  has_many :message_context_statuses
+
+  # Virtual Attributes
+  attr_accessor :email_confirmation, :password, :password_confirmation
```
stepchud: the docs for the [attribute method](https://api.rubyonrails.org/classes/ActiveRecord/Attributes/ClassMethods.html#method-i-attribute) say it doesn't need to be backed by the DB. however, I'm not sure that we need these attributes anyways. AFAIK users shouldn't be able to reset their login.gov password from the app, and I don't know about the email_confirmation.
Screenshot 2024-07-22 at 5 18 35 PM

stepchud (approved):

PR #55, opened 2024-07-25
[39] Create Rails Challenge model

bilalhankins:
As a Rails developer, I want to have a Challenge data model to organize features related to challenges.

Acceptance Criteria:

- [x] This model should mirror the existing ChallengeGov.Challenges.Challenge schema on Phoenix and not require any new migration.

Description:
I created the challenge model and added basic tests to the spec. I also created the agency model due to the association with  challenges


Gemfile
```diff
@@ -2,31 +2,33 @@
 
 source "https://rubygems.org"
 
-ruby "3.2.2"
+source 'https://rubygems.org'
+
+ruby '3.2.2'
```
stepchud: do you know how these were changed? it looks like the " was switched to ' but we don't want that. also the `source` line is duplicated. there is a merge conflict on this file that's related. I'll resolve it but just wanted to call out these quotes shouldn't be converted like this.

bilalhankins: I ran rubocop -A to try to fix the lint issues but that affected way more files than i had thought. I can revert those changes

stepchud: ok yeah I think we should revert the ones and only update if it's failing on the code climate build, otherwise we'll end up switching code styles back and forth constantly

stepchud: make sure you pull first, I've already update a few things on the Gemfile

Rakefile
```diff
@@ -3,6 +3,6 @@
 # Add your own tasks in files placed in lib/tasks ending in .rake,
 # for example lib/tasks/capistrano.rake, and they will automatically be available to Rake.
 
-require_relative "config/application"
+require_relative 'config/application'
```
stepchud: won't comment every file, but I think we have this "prefer single quotes" rule turned off on rubocop now and it shouldn't complain or fail the build on code climate if you're up to date with `dev`. can you revert all these changes so we don't have mismatching styles throughout the codebase? let me know if it's still giving you errors on code climate.

stepchud (commented):
I went ahead and reverted the double quote strings, just leaving this comment for historical purposes

stepchud (approved):
looks good to me

PR #73, opened 2024-07-30
Adjust Asset Pipeline

danivovich:
Issues found in deploy to cloud due to conflicting asset pipelines, etc.

Move to propshaft since we don't need the power of sprockets
Simplify asset builds by using  jsbundling and esbuilt and cssbundling with sass node
Also isolate uswds to its own part of the assets folder, with its output configured to place into the builds folder where propshaft can pick it up.

Also validated that everything works with static assets from public/assets after asset:precompile


app/views/sandbox/index.html.erb
```diff
@@ -362,7 +362,7 @@
     <div class="grid-container">
       <div class="usa-footer__logo grid-row grid-gap-2">
         <div class="grid-col-auto">
-          <img class="usa-footer__logo-img" src="/assets/images/logo-img.png" alt="" />
+          <img width="1024" class="usa-footer__logo-img" src="<%= image_path("challenge-logo.svg") %>" alt="" />
```
stepchud, line 365: 🎉

DEVCONFIG.md
```diff
@@ -0,0 +1,69 @@
...
+Once direnv is installed and your shell is restarted, clone the project and `cd` into it. You should see direnv warn about an untrusted `.envrc` file. Allow the file and finish installing dependencies and setting up the application.
+
+1. Allow direnv to use the envrc file `direnv allow`
+
+## Running Locally
+
+1. Install rubygems dependencies with `bundle install`
+1. Install nodejs dependencies `yarn install`
+1. Set up your uswds files in the build directory `npx gulp copyAssets`
+1. Setup the database `rake db:create`, note that postgres must be running for this to work
```
stepchud, line 64: do we need to note the shared DB with Phoenix here? if a user has their DB created and seeded on that repo, they don't need this step

app/assets/config/manifest.js
stepchud, line 1: ig this is only relevant for sprockets?

DEVCONFIG.md
```diff
@@ -0,0 +1,69 @@
...
+
+Once direnv is installed and your shell is restarted, clone the project and `cd` into it. You should see direnv warn about an untrusted `.envrc` file. Allow the file and finish installing dependencies and setting up the application.
+
+1. Allow direnv to use the envrc file `direnv allow`
+
+## Running Locally
+
+1. Install rubygems dependencies with `bundle install`
+1. Install nodejs dependencies `yarn install`
+1. Set up your uswds files in the build directory `npx gulp copyAssets`
```
stepchud, line 63: nice, this saved me

gulpfile.js
stepchud, line 1: 🎉 this also fixes `bin/rake` for me, which was blowing up on the double linker error (not this particular file, the whole PR)

stepchud (approved):
Looks good to me, thanks for the help!

DEVCONFIG.md
danivovich: Probably, but I figured we'd see how things shake out and update this pretty often

app/assets/config/manifest.js
danivovich: yeah

PR #96, opened 2024-07-31
[40] Create Rails SecurityLog model

bilalhankins:
Created rails SecurityLog Model
Added basic spec tests


spec/models/security_log_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+#
+require 'rails_helper'
+
+RSpec.describe SecurityLog do
+  describe 'Security Log validations' do
+    it 'validates presence of action' do
+      security_log = described_class.new(action: nil)
+      expect(security_log).not_to be_valid
+      expect(security_log.errors[:action]).to include("can't be blank")
+    end
```
stepchud: can this use the new shared_example?
```suggestion
    it_behaves_like 'a model with required attributes', [:actoin]
```

spec/support/shared_examples/validations.rb
```diff
@@ -0,0 +1,10 @@
+# spec/support/shared_examples/validations.rb
+RSpec.shared_examples 'a model with required attributes' do |attributes|
```
stepchud, line 2: 👍

app/models/security_log.rb
```diff
@@ -0,0 +1,49 @@
...
+  attribute :target_id, :integer
+  attribute :target_type, :string
+  attribute :target_identifier, :string
+  attribute :logged_at, :datetime
+
+  private
+
+  def set_logged_at
+    self.logged_at ||= DateTime.now
+  end
```
stepchud, line 54: I just reviewed a similar issue on a different PR and found [this SO solution](https://stackoverflow.com/a/13457972). I think we can leverage the `timestamp_attributes_for_create` method like this:
```
  private_class_method
  def self.timestamp_attributes_for_create
    super << 'logged_at' 
  end
```

app/models/security_log.rb
```diff
@@ -0,0 +1,49 @@
...
+#
+class SecurityLog < ApplicationRecord
+  self.table_name = 'security_log'
+
+  belongs_to :originator, class_name: 'User', optional: true
+
+  validates :action, presence: true, inclusion: { in: %w[
+    status_change account_update role_change accessed_site session_duration
+    create read update delete submit renewal_request
+  ] }
```
stepchud: these could be moved to a constant in case they're useful outside the class
```suggestion
  ROLES = %w[
    status_change account_update role_change accessed_site session_duration
    create read update delete submit renewal_request
  ].freeze
  validates :action, presence: true, inclusion: { in: ROLES }
```

spec/models/security_log_spec.rb
bilalhankins: I was getting errors saying when i did a binding pry i believe the issue was with the set_logged_at so i will try it again with the other suggestion you commented

spec/models/security_log_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+#  target_type           :string(255)
+#  target_identifier     :string(255)
+#  logged_at             :datetime         not null
+#  originator_remote_ip  :string(255)
+#
+require 'rails_helper'
+
+RSpec.describe SecurityLog do
+  describe 'Security Log validations' do
+    it_behaves_like 'a model with required attributes', [:action]
```
stepchud, line 23: @bilalhankins I added this line, works fine for me

stepchud (approved):
g2g

PR #99, opened 2024-08-01
[87] add login.gov certs and configuration

stepchud:
Adds some env configuration to allow both local development and theoretically working on the deployed dev environment.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -28,5 +28,17 @@ module RailsNew
 
     # Use the Postgresql-specific syntax for DB dumps
     config.active_record.schema_format = :sql
+
+    # Shared login.gov config with ENV overrides
+    config.login_gov_oidc = {
+      idp_host: ENV.fetch("LOGIN_IDP_HOST", "https://idp.int.identitysandbox.gov"),
+      login_redirect_uri: ENV.fetch("LOGIN_REDIRECT_EVAL_URL"),
+      logout_redirect_uri: ENV.fetch("LOGOUT_REDIRECT_EVAL_URL"),
+      acr_value: "http://idmanagement.gov/ns/assurance/loa/1",
+      client_id: ENV.fetch("LOGIN_CLIENT_ID"), # determines the login.gov IdP application
+      private_key_password: ENV.fetch("LOGIN_PRIVATE_KEY_PASSWORD", nil), # optional
+      public_key_path: ENV.fetch("LOGIN_PUBLIC_KEY_PATH", "config/public.crt"),
+      private_key_path: ENV.fetch("LOGIN_PRIVATE_KEY_PATH", "config/private.pem"),
+    }
   end
 end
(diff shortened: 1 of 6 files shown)


config/application.rb
```diff
@@ -28,5 +28,17 @@ class Application < Rails::Application
 
     # Use the Postgresql-specific syntax for DB dumps
     config.active_record.schema_format = :sql
+
+    # Shared login.gov config with ENV overrides
+    config.login_gov_oidc = {
+      idp_host: ENV.fetch("LOGIN_IDP_HOST", "https://idp.int.identitysandbox.gov"),
+      login_redirect_uri: ENV.fetch("LOGIN_REDIRECT_EVAL_URL", "https://challenge-dev.app.cloud.gov/auth/result"),
+      logout_redirect_uri: ENV.fetch("LOGOUT_REDIRECT_EVAL_URL", "https://challenge-dev.app.cloud.gov/"),
```
stepchud, line 36: when we have shared sessions, these can point to the same place for both platform and portal apps, but for now they each redirect to themselves since there are separate logins.

config/application.rb
```diff
@@ -28,5 +28,17 @@ class Application < Rails::Application
...
+      acr_value: "http://idmanagement.gov/ns/assurance/loa/1",
+      client_id: ENV.fetch("LOGIN_CLIENT_ID", "urn:gov:gsa:openidconnect.profiles:sp:sso:gsa:_client_id"), # default fake ID for CI
```
stepchud, line 38: we could have the env fail if these ENV are missing, but I went with defaults to make CircleCI/test environments happy without adding unnecessary env vars to them.

manifest.yml
```diff
@@ -20,9 +20,9 @@ applications:
     RAILS_LOG_TO_STDOUT: true
     RAILS_SERVE_STATIC_FILES: true
     HOST: challenge-dev.app.cloud.gov
+    LOGIN_CLIENT_ID: urn:gov:gsa:openidconnect.profiles:sp:sso:gsa:challenge_gov_portal_eval_dev
+    LOGIN_IDP_HOST: https://idp.int.identitysandbox.gov
     LOGIN_PRIVATE_KEY_PATH: dev_key.pem
     LOGIN_PUBLIC_KEY_PATH: dev_cert.pem
-    LOGIN_REDIRECT_URL: https://challenge-portal-dev.app.cloud.gov/auth/result
-    LOGIN_IDP_AUTHORIZE_URL: https://idp.int.identitysandbox.gov/openid_connect/authorize
-    LOGIN_TOKEN_ENDPOINT: https://idp.int.identitysandbox.gov/api/openid_connect/token
```
stepchud, line 27: this app calculates the authorize and token endpoints from `LOGIN_IDP_HOST`, which is the base URL. I assume these don't need to be separate hosts.

config/application.rb
danivovich: 👍

config/application.rb
danivovich: 👍

manifest.yml
danivovich: Sounds good, I had just followed the patterns we'd used on the portal, this makes more sense.

danivovich (approved):
Looks like it will work!

PR #115, opened 2024-08-10
[34] Timeout Warning

### LINKED ISSUE #34: Timeout Warning
### User Story
As a site user, in order to make an informed decisions about my session on the site, I would like to be able to know when I'm about to be logged out from anywhere on the site.

**Acceptance criteria:**

- [x] The system displays a 2 minute countdown warning message before the user is logged out due to inactivity on either application, while they are on any part of the site (Pages, Rails or Phoenix)
- [x] The user can extend their session and they remain logged in on both applications
- [x] The user can end their session and they are logged out right away from both applications
- [x] If the user takes no action, the system logs them out in 2 minutes from both applications
  - [x] The user is taken to a login page with a message displaying saying that the user was logged out due to inactivity

### Definition of Done

**Doing (dev team)**
- [x] Code complete 
- [x] Code is organized appropriately
- [x] Any known trade offs are documented in the associated GH issue
- [x] Code is documented, modules, shared functions, etc.
- [x] Automated testing has been added or updated in response to changes in this PR
- [x] The feature is smoke tested to confirm it meets requirements
- [x] Database changes have been peer reviewed for index changes and performance bottlenecks
- [x] PR that changes or adds UI
  - [x] include a screenshot of the WAVE report for the altered pages
  - [x] Confirm changes were validated for mobile responsiveness
- [x] PR approved / Peer reviewed
- [x] Security scans passed
- [x] Automate accessibility tests passed
- [x] Build process and deployment is automated and repeatable
- [x] Feature toggles if appropriate
- [x] Deploy to staging 
- [x] Move card to testing column in the board
---
**Staging**
- [x] Usability testing: mobile and desktop (Tracy or Marni)
- [ ] Accessibility tested (Marni)
  - [ ]  Keyboard navigation
  - [ ]  Focus confirmed
  - [ ] Color contrast compliance
  - [ ] Screen reader testing 
- [x] Cross browser testing - UI rendering is performant on below listed devices/browsers (Tracy or Marni)
  - [x] Windows/Chrome
  - [x] Windows/Edge
  - [x] Mac/Chrome
  - [x] Mac/Safari
  - [x] iOS/Safari
- [x] AC review (Renata)
- [ ] ~Deploy to production (production-like environment for eval capability) (dev team)~
- [x] Move to production column in the board
---
**Production**
- [ ] ~User and security documentation has been reviewed for necessary updates (Renata/Tracy/Dev team)~ 
- [x] PO / PM approved (Jarah or Renata)
- [x] AC is met and it works as expected (Jarah or Renata)
- [x] Move to done column in the board (Jarah or Renata)

cpreisinger:
#34


app/assets/uswds/_uswds-theme.scss
```diff
@@ -7,7 +7,9 @@ in the form $setting: value,
 ----------------------------------------
 */
 
-@use "uswds-core" with (
```
stepchud: why is this commented?

app/views/layouts/application.html.erb
```diff
@@ -11,10 +11,20 @@
     <%= javascript_include_tag 'application' %>
     <%= javascript_include_tag 'uswds', async: true %>
     <%= javascript_include_tag 'uswds-init', async: true %>
+
+    <script>
+      window.appConfig = {
+        sessionTimeoutMinutes: <%= ENV.fetch('SESSION_TIMEOUT_IN_MINUTES', 15).to_i.minutes %>
+      }
+    </script>
```
stepchud: to keep the application layout a cleaner, could this move to `application.js` with the rest of the timeout code?

stepchud (commented):
forgot to submit this comment earlier

app/controllers/application_controller.rb
```diff
@@ -18,12 +20,27 @@ def sign_in(login_userinfo)
...
   end
 
   def sign_out
     @current_user = nil
     session.delete(:userinfo)
+    session.delete(:session_timeout_at)
+  end
+
+  def renew_session
+    session[:session_timeout_at] = Time.current + ENV.fetch("SESSION_TIMEOUT_IN_MINUTES", 15).to_i.minutes
```
stepchud: I generally try to keep ENV usage to a minimum. Do we keep this in ENV on the Phoenix app? I don't think it needs to change frequently or be configured on different environments. maybe it could be a constant on the controller or LoginGov? not sure the best place, e.g.
```
LoginGov::SESSION_TIMEOUT_IN_MINUTES # or
SessionController::SESSION_TIMEOUT_IN_MINUTES
```

app/javascript/application.js
```diff
@@ -0,0 +1,36 @@
+document.addEventListener("DOMContentLoaded", function () {
+  var sessionTimeoutMinutes = window.appConfig.sessionTimeoutMinutes;
+  var sessionTimeoutMs = sessionTimeoutMinutes * 60 * 1000;
+  var warningTimeoutMs = (sessionTimeoutMinutes - 2) * 60 * 1000;
```
stepchud: these should be declared as `const`

app/javascript/application.js
```diff
@@ -0,0 +1,36 @@
+document.addEventListener("DOMContentLoaded", function () {
+  var sessionTimeoutMinutes = window.appConfig.sessionTimeoutMinutes;
+  var sessionTimeoutMs = sessionTimeoutMinutes * 60 * 1000;
```
stepchud: unused variable
```suggestion
```

app/controllers/application_controller.rb
```diff
@@ -18,12 +20,27 @@ def sign_in(login_userinfo)
...
+  end
+
+  def check_session_expiration
+    if session[:session_timeout_at].present? && session[:session_timeout_at] < Time.current
```
stepchud: if `session[:session_timeout_at].nil?` then shouldn't it also kick the user out? or under what condition would that not be present and is that ok to renew the session if it's not?

stepchud (commented):
nothing major just some small comments or suggestions

stepchud:
I'm seeing the timeout warning even when I'm logged out already. It says your session will expire in without details.

Screenshot 2024-08-14 at 11 00 26 PM

app/views/modals/_renew_session.html.erb
```diff
@@ -0,0 +1,25 @@
...
+  aria-describedby="modal-1-description"
+>
+  <div class="usa-modal__content">
+    <div class="usa-modal__main">
+      <h2 class="usa-modal__heading" id="modal-1-heading">
+        Session expire
+      </h2>
+      <div class="usa-prose">
+        <p id="modal-1-description">
+            Your session will expire in <span id="countdown"></span><br>
```
stepchud: I guess the span doesn't countdown yet

app/controllers/application_controller.rb
cpreisinger: Yeah it exists on the phoenix app. I think initial requirements for the implementation on that end wanted the ability to configure it

.tool-versions
```diff
@@ -1,3 +1,3 @@
-ruby 3.2.4
+ruby 3.2.4
```
stepchud, line 1: whitespace
```suggestion
ruby 3.2.4
```

.envrc
```diff
@@ -5,5 +5,7 @@ use nix
 mkdir -p .nix-bundler
 export BUNDLE_PATH=./.nix-bundler
 
+export SESSION_TIMEOUT_IN_MINUTES=15
```
stepchud: if we decide to keep this as an env var instead of a constant, it needs to be added to the build since `.envrc` won't affect the build.

app/javascript/time_helpers.js
```diff
@@ -0,0 +1,6 @@
+export function formatMilliseconds(ms) {
+  const totalSeconds = Math.round(ms / 1000);
+  const minutes = Math.round(totalSeconds / 60);
```
stepchud: there is a rounding error here. my timeouts are 1 minute and 2 minutes:
```
warningTimeoutMs = 60000;
sessionTimeoutMs = 120000;
```
the countdown timer briefly starts at 1:00, then goes up to 1:59, then right at 1:30 remaining it goes back down to 0:29. Using `Math.floor()` fixes it.
```suggestion
  const minutes = Math.floor(totalSeconds / 60);
```

stepchud (commented):
seems like it is working locally except for the countdown rounding error. it needs some additional configuration for deployment.

app/controllers/application_controller.rb
```diff
@@ -18,12 +20,27 @@ def sign_in(login_userinfo)
...
     session.delete(:userinfo)
+    session.delete(:session_timeout_at)
+  end
+
+  def renew_session
+    session[:session_timeout_at] = Time.current + ENV.fetch("SESSION_TIMEOUT_IN_MINUTES", 15).to_i.minutes
+  end
+
+  def check_session_expiration
+    if logged_in? && (session[:session_timeout_at].blank? || session[:session_timeout_at] < Time.current)
```
stepchud: can it early return when the user is logged out, just to make this logic more readable?
```suggestion
    return unless logged_in?
    
    # user is logged_in
    if session[:session_timeout_at].blank? || session[:session_timeout_at] < Time.current
```

app/views/modals/_renew_session.html.erb
```diff
@@ -0,0 +1,26 @@
+<a href="#renew-modal" style="display:none;" id="renew-modal-open-button" class="usa-button" aria-controls="renew-modal" data-open-modal></a>
+
+<div
```
stepchud, line 3: did you try using the [`<dialog />` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog)? it does nice stuff like input trapping and rendering a background. you can animate the show/hide transition too. you might not need the hidden button to trigger it from javascript if you use the [`dialog.showModal()` function](https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/showModal).

.envrc
```diff
@@ -5,5 +5,7 @@ use nix
...
+
```
stepchud: this can be removed now
```suggestion
```

app/views/modals/_renew_session.html.erb
cpreisinger: I was just basing it off USWDS components. I didn't see dialog mentioned in the docs unless I missed it

https://designsystem.digital.gov/components/modal/

stepchud (changes requested):
Seeing some buggy behavior when I leave the browser open and let it timeout on its own. The countdown keeps going after 0 and never redirects to the logged out page until I refresh the browser manually.

https://github.com/user-attachments/assets/ee321042-97f3-400c-8625-989b7dea3324

stepchud:
> Seeing some buggy behavior when I leave the browser open and let it timeout on its own. The countdown keeps going after 0 and never redirects to the logged out page until I refresh the browser manually.

I logged some console output and it looks like the request to `DELETE /sessions/timeout` happens, which results in a redirect, immediately followed by a `DELETE /dashboard` which is the 404 and for some reason what's logged to the console.
Screenshot 2024-08-23 at 3 44 15 PM

Screenshot 2024-08-23 at 4 22 09 PM
Screenshot 2024-08-23 at 4 22 13 PM

app/controllers/sessions_controller.rb
```diff
@@ -2,6 +2,9 @@
 
 class SessionsController < ApplicationController
   before_action :check_error_result, :require_code_param, :exchange_token, only: [:result]
+  before_action :check_session_expiration, except: [:timeout]
```
stepchud: I was wondering how the inheritance works with the same before_action in ApplicationController and the subcontroller. The Rails guide suggests to use `skip_before_action` here:
```suggestion
  skip_before_action :check_session_expiration, only: [:timeout]
```

stepchud (commented):
this works for me, just a small change on the before_action

stepchud (approved):
looks good to me

PR #187, opened 2024-10-01
[188] Evaluation Form Null Constraint and Validation Shakeup

### LINKED ISSUE #188: Update evaluation_forms model with required data
- [x] Migrate the database for required fields
- [x] Add validations to the Rails model

stonefilipczak:
Brings evaluation forms in line with most recent requirements. 

Includes:

- Migration
- Validations
- Specs for validations 
- Spec helper for creating evaluation forms with valid attributes

### DIFF
--- a/app/models/evaluation_form.rb
+++ b/app/models/evaluation_form.rb
@@ -22,4 +22,9 @@ class EvaluationForm < ApplicationRecord
     joins(challenge: :challenge_manager_users).
       where(challenge_manager_users: { id: user.id })
   }
+
+  validates :title, presence: true
+  validates :instructions, presence: true
+  validates :closing_date, presence: true
+  validates :challenge_phase, presence: true
 end
--- a/app/models/evaluation_form.rb
+++ b/app/models/evaluation_form.rb
@@ -27,4 +27,5 @@ class EvaluationForm < ApplicationRecord
   validates :instructions, presence: true
   validates :closing_date, presence: true
   validates :challenge_phase, presence: true
+  validates :challenge_phase, uniqueness: { scope: :challenge_id }
 end
--- a/app/models/evaluation_form.rb
+++ b/app/models/evaluation_form.rb
@@ -16,7 +16,7 @@
 #  updated_at        :datetime         not null
 #
 class EvaluationForm < ApplicationRecord
-  belongs_to :challenge, optional: true
+  belongs_to :challenge
 
   scope :by_user, lambda { |user|
     joins(challenge: :challenge_manager_users).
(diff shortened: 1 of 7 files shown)


app/models/evaluation_form.rb
stepchud, line 1: can you also remove `optional: true` from belongs_to :challenge on line 19

spec/spec_helper.rb
```diff
@@ -49,6 +49,14 @@ def create_challenge(attrs = {})
   Challenge.create!(user:, agency:, title:, challenge_manager_users:)
 end
 
+def create_evaluation_form(attrs = {})
+  title = attrs[:title] || "test challenge"
+  challenge_id = attrs[:challenge_id] || Challenge.first.id
```
stepchud: this only works if there is already a challenge created in the DB. can you use the create_challenge() helper instead of Challenge.first?
```suggestion
  challenge_id = attrs[:challenge_id] || create_challenge.id
```

db/migrate/20241001143033_change_null_values_on_evaluation_forms.rb
```diff
@@ -0,0 +1,9 @@
+class ChangeNullValuesOnEvaluationForms < ActiveRecord::Migration[7.2]
+  def change
+    change_column_null :evaluation_forms, :title, false
+    change_column_null :evaluation_forms, :instructions, false
+    change_column_null :evaluation_forms, :challenge_phase, false
+    change_column_null :evaluation_forms, :closing_date, false
+    change_column_null :evaluation_forms, :challenge_id, false
```
stepchud, line 7: either in this migration or in a separate one, can we add back the unique index on challenge_id and phase please? you can also add `validates_uniqueness_of :challenge_phase, scope: :challenge_id` to the model to help with error messages

stepchud: oops, my bad! we left that index in the last PR. think we just need the `validates_uniqueness_of` now.

app/models/evaluation_form.rb
```diff
@@ -22,4 +22,9 @@ class EvaluationForm < ApplicationRecord
     joins(challenge: :challenge_manager_users).
       where(challenge_manager_users: { id: user.id })
   }
+
+  validates :title, presence: true
+  validates :instructions, presence: true
+  validates :closing_date, presence: true
+  validates :challenge_phase, presence: true
```
stepchud, line 29: add uniqueness validation

```suggestion
  validates :challenge_phase, presence: true
  validates :challenge_phase, uniqueness: { scope: :challenge_id }
```

stepchud (approved):
niccce 🎉

Issue #160, opened 2024-09-11
Build UI - Eval Form Step 1 & 3 - Evaluation Information


stepchud:
# What/Why
This task will allow users to set the basic details about the Evaluation Form. Step 1 & 3 refer to the mockups:
![Image](https://github.com/user-attachments/assets/b974fba5-008a-4289-965b-0e462732db0a)
![Image](https://github.com/user-attachments/assets/456e2cda-fa81-4133-9013-b35f20b2c8b3)

## Acceptance Criteria
- [x] Implementation matches the mockups, except:
  - [x] Challenge Phase selection should use [USWDS Combo Box](https://designsystem.digital.gov/components/combo-box/) component
- [x] Form data is persisted to the DB
- [x] _Form Title_ is limited to 100 chars
- [x] _Evaluation Instructions_ is limited to 1000 chars
  - [x] The char limit is displayed to the user and counts down to 0 as text is entered
- [x] _Assign to Challenge_ displays all current Challenge Phases for the current user
- [x] Tooltips (?) icons are not clickable for now 
- [x] Form validation happens inline, e.g. if a required field is "skipped" or left blank after being focused there will be a red border on the form element and an error message shown below the field that it is required, same for enforcing character limits if not blocked in the UI

_Disregard the following if this work **does not** include any UI changes_
- [x] WAVE screenshot
- [x] Automated a11y tests (axe-rspec specs)
- [x] Keyboard accessibility (tab navigation, form interactions)
- [x] Responsiveness at desktop and mobile

## Out of Scope
* Tooltips do not need to be implemented/interactable at this time
*  Evaluation Criteria will be handled in other tickets
* Save & Publish will be disabled until the user can input Criteria and the form can be validated

stepchud:
> - [ ] _Form Title_ is limited to XX chars
> - [ ] _Evaluation Instructions_ is limited to XX chars

@r-bartlett-gsa can you provide these limits? It's more a product decision but may need to be informed by design given the placements of form Title in various locations. Feel free to edit the ticket description directly.

stepchud:
@r-bartlett-gsa since the Instructions are required, should we also validate the minimum length or does any single character make it valid? same for Title

r-bartlett-gsa:
@stepchud If there are no technical or database limitations (or design recommendations), I propose the following limits:
Form title: 150
Evaluation instructions: 3000
Criteria title: 150
Criteria description: 1000

cc: @mhotch24

PR #190, opened 2024-10-02
Eval Form Javascript Install / Character Count Implementation

stonefilipczak:
This is a proposal for implementing user feedback on the evaluation form using vanilla JS (as opposed to the previously-discussed stimulus)

The javascript is conditionally inserted into the head of the document when the form is rendered. 

Validating character count for evaluation instructions has been implemented for demonstration purposes. 

I prefer this approach to active record validations on the server side (which was the main value add of using stimulus). Ticket #160 specifies that we want to validate each field one by one as it is unfocused which will be more straightforward to implement using this approach rather than bouncing the entire set of attrs off the server.

updated to use the uswds character count component.

### DIFF
--- a/app/javascript/evaluation_form.js
+++ b/app/javascript/evaluation_form.js
@@ -1,0 +1,5 @@
+document.addEventListener("DOMContentLoaded", function () {
+    document.getElementById("evaluation_form_instructions").addEventListener("input", (e ) => {
+        document.getElementById("character_counter").innerHTML = 3000 - e.target.value.length
+    })
+})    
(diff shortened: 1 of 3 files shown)


app/javascript/evaluation_form.js
stepchud, line 1: don't think we need this now

app/views/evaluation_forms/_form.html.erb
```diff
@@ -33,10 +36,14 @@
         <%= form.text_field :challenge_id %>
       </div>
 
-      <div>
-        <%= form.label :instructions, style: "display: block" %>
-        <%= form.text_field :instructions %>
+      <div class="usa-character-count">
```
stepchud, line 39: looks like this works 👍 can we add the other character counters? in addition to the FE validation, we should probably add the backend schema limits and model validations too.

stepchud (changes requested):

PR #191, opened 2024-10-03
[159] Create EvaluationCriteria model & migration

### LINKED ISSUE #159: Create EvaluationCriteria model & migration
## Description
This model will store the forms that challenge managers create, edit and publish for evaluation of their challenges.

## Attributes
| attribute | type | options |
|--------|--------|--------|
| evaluation_form_id | :id | belongs_to :evaluation_form |
| title | string | null: false |
| description | string | null: false |
| points_or_weight | small_int | null: false |
| scoring_type | enum(:numeric, :rating, :binary) |  default :numeric |
| option_range_start | small_int | default: 0 |
| option_range_start | small_int | default: 4 (? check on this) |
| option_labels | json | [] |

Validations
title, description, evaluation_form_id are required
<evaluation_form_id, title> are unique? (can't have the same title for different criteria on the same form)

Tests
- [x] Add a factory helper for conveniently adding criteria of different kinds to forms.
- [x] Add specs for creating, updating, destroying, validating
- [ ] ~~Criteria for eval forms with weighted scoring must add up to 100~~

cpreisinger:
#159


cpreisinger:
Added missing models for phase
Modified existing models for evaluation_form and user
Added factories for agency, challenge, evaluation_form, phase, and user due to required associations with evaluation_criteria

app/models/evaluation_criterion.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  enum :scoring_type, { numeric: 0, rating: 1, binary: 2 }
+
+  # Validation
+  validates :title, :description, :points_or_weight, presence: true
+  validates :points_or_weight, numericality: { only_integer: true }
+  validates :title,
+            uniqueness: { scope: :evaluation_form_id, message: I18n.t("evaluation_criterion_unique_title_in_form") }
+
+  validate :criteria_weights_sum_to_one_hundred
```
stepchud: I think this should validate at the EvalForm level where, I guess, it will can errors to each of the individual criteria or just one error that the weights don't add up to 100.

db/migrate/20240927010020_create_evaluation_criteria.rb
```diff
@@ -0,0 +1,18 @@
+class CreateEvaluationCriteria < ActiveRecord::Migration[6.1]
+  def change
+    create_table :evaluation_criteria do |t|
+      t.references :evaluation_form, null: false, foreign_key: true
+      t.string :title, null: false
+      t.string :description, null: false
+      t.integer :points_or_weight, null: false
+      t.integer :scoring_type, default: :numeric, null: false 
+      t.integer :option_range_start, default: 0, null: false
+      t.integer :option_range_end, default: 4, null: false
```
stepchud: maybe `option_range_start` and `option_range_stop` should be nullable if the criteria is not a rating scale? I know the ticket says default 0 and 4, but now I'm reconsidering the default should be null. I asked Renata about this on Figma, and she said "there should not be any default options" which I think means even the `scoring_type` would not be selected by default and should error in validation if none is selected.
when we get to the rating scale UI later she wants two options with empty labels in the initial form.

app/models/phase.rb
```diff
@@ -0,0 +1,47 @@
...
+  attribute :end_date, :datetime
+  attribute :open_to_submissions, :boolean
+  attribute :judging_criteria, :text
+  attribute :how_to_enter, :text
+  attribute :challenge_id, :uuid
+
+  # Virtual fields
+  attribute :judging_criteria_length, :integer, default: 0
+  attribute :how_to_enter_length, :integer, default: 0
+  attribute :delete_phase, :boolean, default: false
```
stepchud, line 43: what is this for?

spec/factories/agency.rb
```diff
@@ -0,0 +1,19 @@
+FactoryBot.define do
+  factory :agency do
+    # Associations
+    parent { nil } # Only present on a sub_agency
+
+    # Fields
+    name { "#{Faker::Lorem.word.humanize} Agency" }
```
stepchud, line 7: nice 👍

spec/factories/evaluation_criteria.rb
```diff
@@ -0,0 +1,28 @@
...
+    association :evaluation_form
+
+    # Fields
+    title { "Criterion #{Faker::Lorem.word}" }
+    description { Faker::Lorem.sentence }
+    points_or_weight { rand(0..100) }
+    scoring_type { :numeric }
+    option_range_start { 0 }
+    option_range_end { 4 }
+    option_labels { [] }
```
stepchud: these can be nil for scoring type :numeric

app/models/phase.rb
```diff
@@ -0,0 +1,47 @@
...
+# Table name: phases
+#
+#  id                     :bigint           not null, primary key
+#  uuid                   :uuid             not null, default: -> { "gen_random_uuid()" }
+#  title                  :string
+#  start_date             :datetime
+#  end_date               :datetime
+#  open_to_submissions     :boolean
+#  judging_criteria        :string
+#  judging_criteria_delta  :string
```
stepchud, line 14: more out of curiosity, do you know what the _delta fields are for, or why they're storing the length of the sring in a separate db column?

spec/rails_helper.rb
```diff
@@ -70,4 +71,9 @@
   # config.filter_gems_from_backtrace("gem name")
 
   config.include ActiveSupport::Testing::TimeHelpers
+
+  config.include FactoryBot::Syntax::Methods
+  config.before(:suite) do
+    FactoryBot.reload
```
stepchud, line 77: what does this do?

app/models/phase.rb
cpreisinger: These were virtual fields from the phoenix app. Probably not needed here so I can remove them. 

`delete_phase` specifically was used on submission of that step in the challenge wizard to delete any phases with that delete checkbox checked. Probably somewhat vestigial and can might be or could be done in a better way on the phoenix app.

app/models/evaluation_criterion.rb
cpreisinger: Yeah I agree. I'll get rid of it here and we can do it at the form level so all the weights can be checked on submit. And then maybe later on blur

spec/factories/evaluation_criteria.rb
cpreisinger: The labels? Yeah I need to think of how these will actually be structured. Are these only for `binary` scale? In which case I assume the `option_range_start` and `option_range_end` would be `nil`?

I'm assuming it'd be something like this for example. `[{"disagree": 0, "slightly disagree": 1, .... "agree": 4}]`

The validations on binary scale could get tricky especially taking scoring into account if they aren't validated to be specifically a range under the hood

app/models/phase.rb
cpreisinger: These were used with the rich text editor on the challenge wizard. It was what was used to handle the conversion of the rich text to html.

This is what was used at the time https://quilljs.com/docs/delta

These might eventually be used on the rails app unless we find an alternate way to do the rich text

spec/rails_helper.rb
cpreisinger: Without this line FactoryBot was complaining when I ran the specs. Like it didn't properly load all the factories. I could've also had something else configured wrong or my machine was in a weird state

app/models/phase.rb
stepchud: oh yeah, I haven't even tried setting up multiple phases on a challenge yet 🙃

spec/factories/evaluation_criteria.rb
stepchud: oh sorry, labels can probably default to empty list [] just so it's valid json but i meant more the range_start and range_end. labels only apply for rating and binary scale but it can be empty for numeric scale. 👌

stepchud: for binary scale scoring it's all or nothing scoring, so the submission either gets the full points for that criteria or 0 depending on the judges choice on the binary score.

db/migrate/20240927010020_create_evaluation_criteria.rb
cpreisinger: Ok yeah I can remove the defaults and allow null

spec/factories/evaluation_criteria.rb
cpreisinger: Oh that kind of binary. So does this often end up being a boolean type of yes/no? Are there cases where they'd label them as something else?

cpreisinger: I see they can add a 0 and 1 label. So that's fine

stepchud: yes binary scale has customizable labels, I guess so it can support yes/no, pass/fail, and I don't know what else 🤔

app/models/evaluation_form.rb
```diff
@@ -17,6 +17,7 @@
 #
 class EvaluationForm < ApplicationRecord
   belongs_to :challenge
+  has_many :evaluation_criteria, dependent: nil
```
stepchud: this should be dependent: :destroy so the associated criteria are destroyed when the form is destroyed
```suggestion
  has_many :evaluation_criteria, dependent: :destroy
```

stepchud (approved):
LGTM

PR #214, opened 2024-10-14
[199] Create Evaluator Invitation and Challenge Phases Evaluators

### LINKED ISSUE #199: Create EvaluatorInvitation model(s)
# What/Why
This task adds a temporary record for tracking Evaluators that have been invited to participate but have not yet setup an account on login.gov. This is necessary because we don't have all required information about the User when the CM invites them, until they create an account and login for the first time.

We also need a table to join Evaluator users with their assigned challenge phase.

## Attributes
**Table name** `evaluator_invitations`
| attribute | type | options |
|--------|--------|--------|
| challenge_id | :id | belongs_to :challenge |
| phase_id | :id | belongs_to :phase |
| first_name | string | null: false |
| last_name | string | null: false |
| email | string | null: false |
| last_invite_sent | datetime | null: false |
| timestamps |

Unique index on challenge_id / phase_id / email so that only one invite can be sent per challenge/phase/email combination.

**Table name** `challenge_phases_evaluators`
| attribute | type | options |
|--------|--------|--------|
| challenge_id | :id | belongs_to :challenge |
| phase_id | :id | belongs_to :phase |
| user_id | :id | belongs_to :user |
| timestamps |

## Acceptance Criteria
- [x] A requirement for completion
  - [x] Sub-requirement (optional)
  - Additional notes (optional)

**There are no UI changes in this ticket.**
- [ ] ~WAVE screenshot~
- [ ] ~Automated a11y tests (axe-rspec specs)~
- [ ] ~Keyboard accessibility (tab navigation, form interactions)~
- [ ] ~Responsiveness at desktop and mobile~

emmabjj:
[#199](https://github.com/GSA/Challenge_platform/issues/199)

For PR adds the migrations, models and test factories for `Evaluator Invitation` and the join table `Challenge Phases Evaluators`. 

This task adds a temporary record for tracking Evaluators that have been invited to participate but have not yet setup an account on login.gov. This is necessary because we don't have all required information about the User when the CM invites them, until they create an account and login for the first time.

Challenge Phases Evaluators joins Evaluator users with their assigned challenge phase.

No UI changes in this ticket.


app/models/user.rb
```diff
@@ -48,6 +48,8 @@ class User < ApplicationRecord
            dependent: :destroy
   has_many :submission_documents, class_name: 'Submissions::Document', dependent: :destroy
   has_many :message_context_statuses, dependent: :destroy
+  has_many :challenge_phases_evaluators, dependent: :destroy
+  has_many :evaluated_challenges, through: :challenge_phases_evaluators, source: :challenge
```
stepchud: this is a valid relationship but I think the phases might be more useful since they will be unique while there could be duplicate challenges per user.

db/migrate/20241014210800_create_evaluator_invitations.rb
```diff
@@ -0,0 +1,16 @@
+class CreateEvaluatorInvitations < ActiveRecord::Migration[7.2]
+  def change
+    create_table :evaluator_invitations do |t|
+      t.references :challenge, null: false, foreign_key: true
+      t.references :phase, null: false, foreign_key: true
+      t.string :first_name, null: false
+      t.string :last_name, null: false
+      t.string :email, null: false
+      t.datetime :last_invite_sent, null: false
```
stepchud: I was wrong, `last_invite_sent` is nullable. we can't send the invites yet and even when we do they will be a delayed job so the record will be created before the invite is sent and then updated when the job is processed.

stepchud (commented):
There might be a few specs just to exercise creating and destroying EvaluatorInvitation and ChallengePhaseEvaluator records?

emmabjj:
@stepchud Thanks for the feedback, I pushed those fixes and added tests. Looks like there's something going on with the how the Code Climate Test Reporter ID is set.

stepchud (approved):
looks good!

emmabjj:
> looks good!

Great, I'll go ahead and merge it - thanks!

PR #221, opened 2024-10-21
Eval Form Datepicker Fix

stonefilipczak:
The datepicker for closing date on the eval form was broken because the combo box element was throwing a javascript error which caused all of the USWDS javascript to crash. 

Fixed this by switching the name attribute of the combo box select to be an id attribute. This wouldn't ordinarily be required in HTML but is needed for how USWDS javascript strips down and rebuilds the element on initialization. 

This further necessitated the introduction of a title attribute which satisfies the wave tool.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,7 +71,7 @@ jobs:
   rspec:
     executor:
       name: test_executor
-    
+
     environment:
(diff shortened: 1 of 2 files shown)


stepchud (approved):
👍

PR #254, opened 2024-11-04
[245] Invited Evaluators User Setup

### LINKED ISSUE #245: Invited evaluators' user setup requirements
# What/Why
As a Challenge Manager, I want to invite users to be evaluators of my challenge. Currently when I invite a user who has not registered with challenge.gov yet (no User exists), and they login with a valid login.gov account, they are created with the `solver` role. If there is an open invitation for that user's email, they should be created with `evaluator` role instead.

## Acceptance Criteria
A user who creates a new account after being invited:
- [x] has the `evaluator` role.
- [x] associated `ChallengePhasesEvaluator` record for the invited Phase 
- [x] Invitation record is deleted
- [x] Initial status is `pending`
  - [x] `pending` status does not have access to the app features
- [x] An admin can set them to `active` status (on the Phoenix app)
  - [x] When they are `active`, they can use evaluator features and functionality

- [ ] ~WAVE screenshot~
- [ ] ~Automated a11y tests (axe-rspec specs)~
- [ ] ~Keyboard accessibility (tab navigation, form interactions)~
- [ ] ~Responsiveness at desktop and mobile~

emmabjj:
#245 

Update the invited evaluator's user setup requirements:
- A user with a new account defaults to role of evaluator with status pending. 
- An evaluator is associated with the phase they are assigned through `ChallengePhasesEvaluator`
- When an new evaluator user is successfully registered, their `EvaluatorInvitation` record is deleted and they are then associated to the Phase they successfully registered for through `ChallengePhasesEvaluator`
- An admin can change a new evaluator user's status to `active`

Tests added to cover changes.

No UI changes/No WAVE screenshot

### DIFF
--- a/app/controllers/manage_evaluators_controller.rb
+++ b/app/controllers/manage_evaluators_controller.rb
@@ -78,13 +78,15 @@ class ManageEvaluatorsController < ApplicationController
 
   # Create action helpers
   def process_evaluator_invitation(email)
-    existing_invitation = @challenge.evaluator_invitations.find_by(email:, phase: @phase)
-    user = User.find_by(email:)
+    user = User.find_by(email: email)
+    existing_invitation = @challenge.evaluator_invitations.find_by(email: email, phase: @phase)
 
-    if existing_invitation
+    if user
+      result = add_user_as_evaluator(user)
+      existing_invitation&.destroy if result[:success]
+      result
+    elsif existing_invitation
       resend_invitation(existing_invitation)
-    elsif user && valid_evaluator_role?(user)
-      add_user_as_evaluator(user)
     else
       create_new_invitation(email)
     end
@@ -105,8 +107,11 @@ class ManageEvaluatorsController < ApplicationController
   end
 
   def add_user_as_evaluator(user)
-    cpe = ChallengePhasesEvaluator.find_or_create_by(challenge: @challenge, phase: @phase, user:)
+    cpe = ChallengePhasesEvaluator.find_or_create_by(challenge: @challenge, phase: @phase, user: user)
+    invitation = @challenge.evaluator_invitations.find_by(email: user.email, phase: @phase)
+
     if cpe.persisted?
+      invitation&.destroy
       { success: true, message: "#{user.email} has been added as an evaluator for this phase." }
     else
       { success: false, message: "Failed to add #{user.email} as an evaluator." }
--- a/app/controllers/manage_evaluators_controller.rb
+++ b/app/controllers/manage_evaluators_controller.rb
@@ -179,6 +179,7 @@ class ManageEvaluatorsController < ApplicationController
 
   def render_json_response(result)
     if result[:success]
+      flash[:notice] = result[:message]
       render json: { success: true, message: result[:message] }
     else
       render json: { success: false, message: result[:message] }, status: :unprocessable_entity
--- a/app/controllers/manage_evaluators_controller.rb
+++ b/app/controllers/manage_evaluators_controller.rb
@@ -6,8 +6,6 @@ class ManageEvaluatorsController < ApplicationController
   before_action :set_challenge
   before_action :set_phases, only: [:index]
 
-  VALID_EVALUATOR_ROLES = %w[evaluator solver challenge_manager].freeze
-
   def index
     if @phases.empty?
       handle_empty_phases
@@ -18,12 +16,12 @@ class ManageEvaluatorsController < ApplicationController
 
   def create
     @phase = @challenge.phases.find(evaluator_invitation_params[:phase_id])
-    result = process_evaluator_invitation(evaluator_invitation_params[:email])
+    user = User.find_by(email: evaluator_invitation_params[:email])
 
-    if result[:success]
-      handle_successful_creation(result)
+    if existing_evaluator?(user)
+      handle_existing_evaluator(user)
     else
-      handle_failed_creation
+      process_new_evaluator
     end
   end
 
@@ -45,6 +43,10 @@ class ManageEvaluatorsController < ApplicationController
     @phases = @challenge.phases.order(:start_date)
   end
 
+  def existing_evaluator?(user)
+    user && ChallengePhasesEvaluator.exists?(challenge: @challenge, phase: @phase, user: user)
+  end
+
   def evaluator_invitation_params
     params.require(:evaluator_invitation).permit(
       :first_name, :last_name, :email, :challenge_id, :phase_id, :last_invite_sent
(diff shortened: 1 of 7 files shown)


app/controllers/manage_evaluators_controller.rb
```diff
@@ -78,13 +78,15 @@ def fetch_existing_evaluators
...
-
-    if existing_invitation
+    user = User.find_by(email: email)
+    existing_invitation = @challenge.evaluator_invitations.find_by(email: email, phase: @phase)
+
+    if user
+      result = add_user_as_evaluator(user)
+      existing_invitation&.destroy if result[:success]
+      result
+    elsif existing_invitation
```
stepchud: You're finding and destroying the EvaluatorInvitation record twice here, which should probably be consolidated to conditionally destroy it in `add_user_as_evaluator`

```suggestion
    user = User.find_by(email: email)
    existing_invitation = @challenge.evaluator_invitations.find_by(email: email, phase: @phase)

    if user
      add_user_as_evaluator(user, existing_invitation)
    elsif existing_invitation
```

app/models/user.rb
```diff
@@ -133,7 +133,7 @@ def self.default_role_and_status_for_email(email)
     if default_challenge_manager?(email)
       %w[challenge_manager pending]
     else
-      %w[solver active]
+      %w[evaluator pending]
     end
```
stepchud: This logic needs to be modified a bit. If a new user has an existing EvaluatorInvitation, their role and status should be `[evaluator, pending]`, otherwise it can use the existing `default_challenge_manager?(email)` condition and default to `[solver, active]` like it currently does.
Additionally, we need to add an [after_create callback](https://guides.rubyonrails.org/active_record_callbacks.html#create-callbacks) on User to deal with the Invitation. This will only be called the first time the User record is created and it can handle multiple invitations to different challenges.
```
class User
  after_create :accept_evaluator_invitation
  
  private
  def accept_evaluator_invitation
    EvaluatorInvitation.where(email: self.email).each do |invite|
      ChallengePhasesEvaluator.create(challenge: invite.challenge, phase: invite.phase, user: self)
      invite.destroy
    end
  end
end
```

emmabjj: @stepchud Okay I understand, I'll modify the logic here and add some tests to cover this - thanks!

app/controllers/manage_evaluators_controller.rb
```diff
@@ -127,12 +117,27 @@ def handle_successful_creation(result)
...
-  def handle_failed_creation
-    @evaluator_invitations = fetch_evaluator_invitations
-    @existing_evaluators = fetch_existing_evaluators
+  def handle_failed_creation(error_message)
+    flash.now[:alert] = error_message
+    fetch_evaluators_and_invitations
     render :index
   end
 
+  # prevent duplicate evaluators or evauator invitations
```
stepchud: typo
```suggestion
  # prevent duplicate evaluators or evaluator invitations
```

app/models/challenge_phases_evaluator.rb
```diff
@@ -15,4 +15,14 @@ class ChallengePhasesEvaluator < ApplicationRecord
   belongs_to :challenge
   belongs_to :phase
   belongs_to :user
+
+  validate :user_has_valid_role, if: -> { user.present? }
```
stepchud: `challenge_id` `phase_id` and `user_id` are all required (`NOT NULL`) fields. Rails also validates the presence of `belongs_to` relationships (unless you declare them with `optional: true`). so basically, `user.present?` check is not needed.

spec/controllers/manage_evaluators_controller_spec.rb
```diff
@@ -84,27 +84,91 @@
...
-            last_invite_sent: Time.current
+    context 'when inviting an existing user with an invitation' do
+      let(:existing_user) { create(:user, role: 'evaluator', status: 'pending') }
+      let!(:invitation) { create(:evaluator_invitation, challenge: challenge, phase: phase, email: existing_user.email) }
+
+      before do
+        ChallengePhasesEvaluator.create!(challenge: challenge, phase: phase, user: existing_user)
+      end
+
+      it 'adds the user as an evaluator without creating a new ChallengePhasesEvaluator' do
```
stepchud: I'm confused which scenario is being tested here. I don't think it should be possible for an existing user to also have an invitation? If an existing user is invited to the challenge, then a ChallengePhaseEvaluator will be created for them and the Invitation should be deleted.. might need to add that delete logic to the controller actually.

the use of `let`, `let!` and `before` blocks makes it hard for me to follow the logical ordering of events in this test. I think this test should be something like "it does not create an EvaluatorInvitation or ChallengePhaseEvaluator when the user is already added to the challenge" and then you can remove the `let!(:invitation)` since there should not be an invitation for an existing user possible.

spec/controllers/manage_evaluators_controller_spec.rb
```diff
@@ -84,27 +84,91 @@
...
+        expect(flash[:notice]).to include("has been added as an evaluator for this phase")
+      end
+    end
+
+    context 'when adding an existing user with an invalid role' do
+      let(:existing_user) { create(:user, role: 'evaluator', status: 'pending') }
+
+      before do
+        existing_user.update_column(:role, 'admin') # invalid evaluator role
+      end
```
stepchud: can't you create the user as an admin instead of updating them after create?
```suggestion
      let(:existing_user) { create(:user, role: 'admin', status: 'active') }
```

spec/models/challenge_phases_evaluator_spec.rb
```diff
@@ -62,4 +62,29 @@
     expect(phase.evaluators).to include(user)
     expect(user.evaluated_phases).to include(phase)
   end
+
+  context "with invalid user role" do
+    let(:invalid_user) { create(:user, role: User::VALID_EVALUATOR_ROLES.first) }
+
+    before do
+      invalid_user.update_column(:role, 'admin')
```
stepchud: seems simpler to create the user with the role you want instead of updating them after create

spec/models/challenge_phases_evaluator_spec.rb
```diff
@@ -62,4 +62,29 @@
...
+    end
+
+    it "is invalid" do
+      evaluator = build(:challenge_phases_evaluator, challenge:, phase:, user: invalid_user)
+      expect(evaluator).not_to be_valid
+      expect(evaluator.errors[:user]).to include("must have a valid evaluator role")
+    end
+  end
+
+  User::VALID_EVALUATOR_ROLES.each do |role|
```
stepchud, line 58: 👍

spec/models/user_spec.rb
```diff
@@ -174,4 +174,80 @@
...
+    let(:user_email) { 'test@example.com' }
+
+    context 'when there are existing evaluator invitations' do
+      before do
+        create(:evaluator_invitation, challenge: challenge1, phase: phase1, email: user_email)
+        create(:evaluator_invitation, challenge: challenge1, phase: phase2, email: user_email)
+        create(:evaluator_invitation, challenge: challenge2, phase: phase3, email: user_email)
+      end
+
+      it 'creates ChallengePhasesEvaluator records for all invitations when user is created' do
```
stepchud, line 193: 👍

spec/models/user_spec.rb
```diff
@@ -174,4 +174,80 @@
...
+        }.to change(ChallengePhasesEvaluator, :count).by(3)
+      end
+
+      it 'destroys all EvaluatorInvitation records when user is created' do
+        expect {
+          create(:user, email: user_email, role: 'evaluator')
+        }.to change(EvaluatorInvitation, :count).by(-3)
+      end
+
+      it 'associates the new user with the correct challenges and phases' do
```
stepchud, line 205: 💯

spec/models/user_spec.rb
```diff
@@ -174,4 +174,80 @@
...
+      end
+
+      it 'does not destroy any EvaluatorInvitation records' do
+        expect {
+          create(:user, email: user_email)
+        }.not_to change(EvaluatorInvitation, :count)
+      end
+    end
+
+    context 'when there are invitations for multiple challenges and phases' do
```
stepchud: are these duplicates of the first context 'when there are existing evaluator invitations'? seems like they're testing the same things

stepchud (changes requested):
review feedback

spec/controllers/manage_evaluators_controller_spec.rb
emmabjj: Yeah, I split this logic out - this test is supposed to make sure that an additional ChallengePhaseEvaluator isn't created when that evaluator is already associated with that challenge/phase. I added the same test for an EvaluatorInvitation - an additional EvaluatorInvitation isn't created when an previous invitation for that evaluator and challenge/phase exists. 

Do I need to add the delete logic to the controller if the deleting the invitation is handled as an after_create on the user already? 

```
class User
  after_create :accept_evaluator_invitation
  
  private
  def accept_evaluator_invitation
    EvaluatorInvitation.where(email: self.email).each do |invite|
      ChallengePhasesEvaluator.create(challenge: invite.challenge, phase: invite.phase, user: self)
      invite.destroy
    end
  end
end
```

spec/models/user_spec.rb
emmabjj: Yep, I think that test was leftover before I decided to add challenges with multiple phases - thanks for pointing that out!

stepchud (approved):

spec/controllers/manage_evaluators_controller_spec.rb
stepchud: I think you're right 👍

PR #279, opened 2024-11-19
[200] Update the Invitation routes

stepchud:
This PR has mostly the routes and controller/resource changes that go along with them.
Some feedback/suggestions for `200_manage_evaluators_list` see the comments for details.


app/controllers/evaluator_invitations_controller.rb
stepchud, line 1: Moved `/evaluator_invitations/:id/resend` => `/phases/:phase_id/evaluators/:id/resend_invite`
Just consolidating this since it's the only action on this controller, is a non-REST route, and can exist on the nested route for EvaluatorsController.

app/controllers/evaluators_controller.rb
```diff
@@ -1,13 +1,14 @@
 # frozen_string_literal: true
 
-class ManageEvaluatorsController < ApplicationController
-  include ManageEvaluatorsHelper
```
stepchud, line 4: helpers are automatically included in Rails controllers/views, so this isn't necessary

app/controllers/evaluators_controller.rb
```diff
@@ -1,13 +1,14 @@
...
-
+class EvaluatorsController < ApplicationController
   before_action -> { authorize_user('challenge_manager') }
-  before_action :set_phase, only: [:index, :create, :destroy]
+
+  # All routes are scoped to a Challenge Phase
+  before_action :set_challenge_phase
```
stepchud, line 7: relies on `phase_id` param from the nested route definition:
`/phases/:phase_id/evaluators/*`

app/controllers/evaluators_controller.rb
```diff
@@ -1,13 +1,14 @@
...
 
   def index
-    handle_existing_phases
+    @evaluator_invitations = @phase.evaluator_invitations
+    @existing_evaluators = @phase.evaluators
```
stepchud, line 11: I think it's easier to read directly inside the action when there is some specific data needed

app/controllers/evaluators_controller.rb
```diff
@@ -21,14 +22,30 @@ def create
...
   def destroy
     result = evaluator_service.remove_evaluator(params[:evaluator_type], params[:id])
-    render_json_response(result)
+
+    if result[:success]
+      flash[:notice] = result[:message]
+      render json: { success: true, message: result[:message] }
+    else
+      render json: { success: false, message: result[:message] }, status: :unprocessable_entity
+    end
```
stepchud, line 31: inlining this so it's easier to see what is rendered by the `destroy` action

app/controllers/evaluators_controller.rb
```diff
@@ -21,14 +22,30 @@ def create
...
+  end
+
+  def resend_invite
```
stepchud, line 34: the new route that was copied from `EvaluatorInvitationsController`

app/controllers/evaluators_controller.rb
```diff
@@ -21,14 +22,30 @@ def create
...
+    @evaluator_invitation = @phase.evaluator_invitations.find(params[:id])
+    if evaluator_service.resend_invitation(@evaluator_invitation)
```
stepchud: use the service in the controller when possible

app/javascript/controllers/delete_evaluator_modal_controller.js
```diff
@@ -43,30 +43,29 @@ export default class extends Controller {
...
     })
     .then(response => {
-      if (!response.ok) {
-        throw new Error('Network response was not ok')
-      }
-      const data = response.json()
+      if (!response.ok) { throw new Error('Network response was not ok') }
+      return response.json()
+    })
+    .then(data => {
```
stepchud, line 55: I broke this on your PR, the DELETE request is working but the response wasn't being handled correctly with promises so I added this back, sorry!

app/services/evaluator_management_service.rb
```diff
@@ -28,7 +28,17 @@ def self.accept_evaluator_invitation(user)
       ChallengePhasesEvaluator.create(challenge: invite.challenge, phase: invite.phase, user:)
       invite.destroy
     end
-    { success: true, message: I18n.t('manage_evaluators.accept_evaluator_invitation.success') }
+    { success: true, message: I18n.t('evaluators.accept_evaluator_invitation.success') }
+  end
+
+  # TODO: Implement sending the actual invitation email here
+  def resend_invitation(invitation)
```
stepchud, line 35: make it non-private for reusability

app/views/evaluators/_delete_evaluator_modal.html.erb
```diff
@@ -34,15 +34,5 @@
         </ul>
       </div>
     </div>
-    <button
```
stepchud, line 37: there's no close icon in Figma and this markup was erroring out because the sprite.svg icon was not found so I just removed it here. user can still escape/close the modal without the X icon button.

spec/requests/evaluators_spec.rb
```diff
@@ -1,37 +1,41 @@
...
+    it 'renders invitations and evaluators' do
+      invitation # create existing invitation
+      associate_challenge_manager_challenge
+      associate_evaluator_challenge_phase
+      get phase_evaluators_path(phase)
 
-      expect(assigns(:evaluator_invitations)).to eq([invitation])
-      expect(assigns(:existing_evaluators)).to eq([evaluator])
+      expect(response.body).to include(invitation.email)
+      expect(response.body).to include(evaluator.email)
```
stepchud, line 28: expect content in the request response instead of testing assignment of instance variables, which is an implementation detail that could change in the future.

spec/requests/evaluators_spec.rb
```diff
@@ -1,37 +1,41 @@
...
-  end
+  let(:evaluator_service_double) { instance_double(EvaluatorManagementService) }
+  let(:evaluator_service) { EvaluatorManagementService.new(challenge, phase) }
 
   describe 'GET #index' do
     before do
-      invitation # create existing invitation
-      challenge.challenge_phases_evaluators.create(user: evaluator, phase: phase)
+      mock_evaluator_service
+      login_challenge_manager
```
stepchud, line 18: trying to be more explicit about how each of these describe blocks are setup

spec/requests/evaluators_spec.rb
```diff
@@ -41,17 +45,17 @@
       end
 
       it 'calls the EvaluatorManagementService to process the invitation' do
-        expect(evaluator_service).to receive(:process_evaluator_invitation).and_return({ success: true,
-                                                                                         message: 'Invitation sent successfully.' })
-        post phase_manage_evaluators_path(phase), params: valid_params
-        expect(response).to redirect_to(phase_manage_evaluators_path(phase))
+        expect(evaluator_service_double).to receive(:process_evaluator_invitation).and_return({ success: true,
```
stepchud, line 48: I wasn't sure about creating this test double for most of these specs but I left it in anyways. was there a reason you wanted to mock this always in the request specs?

emmabjj: Good question - yeah just separating concerns, by mocking the service here I wanted to ensure the request specs were able to isolate and test the controller's behavior and response handling, rather than only testing the business logic in the service - which then is tested in the `spec/services/evaluator_management_service_spec.rb` anyways.

app/views/evaluators/_delete_evaluator_modal.html.erb
emmabjj: Yeah I noticed that, so in the js I had added modal close on click outside. Makes sense to remove this - it was part of the modal template on the uswds site. 👍

emmabjj (approved):
These changes make sense to me - thanks @stepchud

Issue #209, opened 2024-10-10
Create Total Submissions stats component


stepchud:
# What/Why
The stats section at the top of the submission list view. These stats could be expensive to query so let's consider performance  implications while designing this.

[[Story](https://github.com/GSA/Challenge_platform/issues/79)]

### Figma Link
[stats section](The stats section at the top could be expensive to query and we may want to cache this component since it doesn't change that frequently and slightly out of date stats are okay.)

## Acceptance Criteria
- [ ] uses USWDS [Summary box](https://designsystem.digital.gov/components/summary-box/) component, perhaps?

_Disregard the following if this work **does not** include any UI changes_
- [ ] WAVE screenshot
- [ ] Automated a11y tests (axe-rspec specs)
- [ ] Keyboard accessibility (tab navigation, form interactions)
- [ ] Responsiveness at desktop and mobile

PR #280, opened 2024-11-20
[209] Submission Stats

stonefilipczak:
https://github.com/GSA/Challenge_platform/issues/209
Adds a stats component to the submissions list page for challenge managers. 

Screen Shot 2024-11-19 at 10 10 23 PM


stepchud:
This will be updated with status from the `Evaluation` records, which are currently in review on #281

app/views/phases/_submissions_stats.html.erb
```diff
@@ -0,0 +1,66 @@
...
+    </div>
+    <div class="display-flex flex-wrap tablet:flex-justify margin-top-2">
+        <div class="font-body-xs">
+            <%= image_tag(
+                "images/usa-icons/verified.svg",
+                class: "usa-icon--size-3 margin-right-1",
+                alt: ""
+            )%>
+            <span class="text-top">
+                <span class="text-primary"><%= @submissions.select { |s| s.eligible_for_evaluation? }.count %> of <%= @submissions.count %></span>
```
stepchud: prefer queries to memory counting

app/views/phases/_submissions_stats.html.erb
```diff
@@ -0,0 +1,66 @@
...
+            </span>
+        </div>
+        <div class="font-body-xs">
+            <%= image_tag(
+                "images/usa-icons/star_outline.svg",
+                class: "usa-icon--size-3 margin-right-1",
+                alt: ""
+            )%>
+            <span class="text-top">
+                <span class="text-primary"><%= @submissions.select { |s| s.selected_to_advance? }.count %> of <%= @submissions.count %></span>
```
stepchud: prefer queries to memory counting

app/views/phases/submissions.html.erb
```diff
@@ -6,5 +6,6 @@
     <p>This challenge phase does not currently have any submissions.</p>
   </div>
 <% else %>
+  <%= render partial: "submissions_stats", locals: { submissions: @submissions } %>
```
stepchud: don't think the locals are needed here (or below) since we already have the instance variable defined and available to the views

app/views/phases/_submissions_stats.html.erb
```diff
@@ -0,0 +1,66 @@
...
+            <div class="text-center margin-right-5">
+            <span class="font-sans-xl text-green text-bold"><%= @submissions_by_status["completed"] || 0 %></span><br>
+            <span class="text-base-sm text-green text-bold">Completed</span>
+            </div>
+            <div class="text-center margin-right-5">
+            <span class="font-sans-xl text-orange text-bold"><%= @submissions_by_status["in_progress"] || 0 %></span><br>
+            <span class="text-base-sm text-orange text-bold">In Progress</span>
+            </div>
+            <div class="text-center margin-right-2">
+            <span class="font-sans-xl text-secondary-dark text-bold"><%= @submissions_by_status["not_started"] || 0 %></span><br>
```
stepchud: update statuses

stepchud (commented):
First review pass

stepchud (changes requested):
(I think) I made the necessary code changes as far as the stats component calculations and display. However, I would like to see some tests for this component. They should mostly be request specs in `spec/requests/submissions_spec.rb` that assert the various stats are displayed correctly.  Can we also add a `spec/system/submissions_spec.rb` expectation that the stats show visible content on the page.

app/views/phases/_submissions_stats.html.erb
```diff
@@ -15,7 +15,7 @@
                 Total Submissions
                 </h3>
                 <p class="usa-summary-box__text text-primary-darker text-bold">
-                Evaluations due by <%= @phase.evaluation_form.closing_date.strftime('%m/%d/%Y') %>
+                Evaluations due by <%= if @phase.evaluation_form then @phase.evaluation_form.closing_date.strftime('%m/%d/%Y') end %>
```
stepchud: I've rarely seen the `if .. then .. end` conditional form, there are a few different ways to write this that are a little more concise. It should fallback to 'N/A' if there is no form too. imo the [safe-navigation operator](https://ruby-doc.org/core-2.6/doc/syntax/calling_methods_rdoc.html#label-Safe+navigation+operator) works better here..
```suggestion
                Evaluations due by <%= @phase.evaluation_form&.closing_date&.strftime('%m/%d/%Y') || "N/A" %>
```

stepchud: also this should probably be moved to a helper function, to help reuse the logic, since we use the same date a few places.

spec/requests/submissions_spec.rb
```diff
@@ -53,6 +53,19 @@
         get submissions_phase_path(phase)
         expect(response).to have_http_status(:not_found)
       end
+
+      it "renders submission statistics" do
+        submission = create(:submission, challenge: challenge, phase: phase)
+        submission_2 = create(:submission, challenge: challenge, phase: phase, judging_status: "selected")
```
stepchud: [Lint/UselessAssignment](https://docs.rubocop.org/rubocop/cops_lint.html#lintuselessassignment)
```suggestion
        create(:submission, challenge: challenge, phase: phase)
        create(:submission, challenge: challenge, phase: phase, judging_status: "selected")
```

spec/requests/submissions_spec.rb
```diff
@@ -53,6 +53,19 @@
...
+
+
+        get submissions_phase_path(phase)
+        expect(response.body).to include("Boston Tea Party Cleanup")
+        # total submissions
+        expect(response.body).to include("2")
```
stepchud: this is not testing much since the expect below also has a 2 in it, but I found a way to use css selectors that gives more specific expectations. the trick is to `include Capybara::RSpecMatchers` in the request specs to enable the `have_css()` matcher. I'll add this now.

stepchud (approved):
nice work, I added the rspec matcher update e20849c

PR #288, opened 2024-11-25
[275-FIX] Fix for max input values for scale types

### LINKED ISSUE #275: Evaluation Form: Limit criteria points / weight field input to 4 digits
As a user, I want to enter reasonable numbers for criteria points / weight.

Based on [Feedback from QA](https://github.com/GSA/Challenge_platform/issues/69#issuecomment-2462912774), this number is not limited in the UI and it can cause errors. We should restrict the input to 4 digits (<= 9999) and <= 100 for weighted scale.

## Technical Notes
Could use [HTML input max attribute](https://www.w3schools.com/tags/att_input_max.asp) here

## Acceptance Criteria
- [x] User can not enter more than 4 digits for criteria points
- [x] Scores for weighted (%) criteria must be <= 100
- [x] When the user switches between Scale Types, the form should remain interactive and enable the user to fix any validation issues and make changes to the form as necessary (see [bug report](https://github.com/GSA/Challenge_platform/issues/275#issuecomment-2480005058) below)

- [x] WAVE screenshot

cpreisinger:
For this PR I ended up expanding all the accordions and just blanking all points/weights if two conditions were met
- The user was swapping to a weighted scale type
- The user had at least one criteria where the points/weight was > 100
A confirmation alert also shows up currently and prevents scale type switching if it isn't confirmed

An additional update on this PR was changing the way the inputs worked for points/weight to validate their length on blur using `reportValidity` HTML5 validations. This was to prevent being able to temporarily circumvent min and max values by typing

Possible alteration to make it more inline with the initial comment if we want would be:
- Make it always expand all criteria after a scale change (this would allow normal HTML5 validations to trigger)
- Don't blank out values

#275


app/javascript/controllers/evaluation_criteria_controller.js
```diff
@@ -100,6 +100,18 @@ export default class extends Controller {
...
   }
 
+  checkPointsOrWeightMax(event) {
+    const input = event.target;
+    const min = input.min;
+    const max = input.max;
+    const value = input.value;
+
+    // If field has a value on blur then check if it's valid
+    if ((value && value < min) || value > max) {
```
stepchud: I'm not sure how to test this but shouldn't the parens surround the min/max?
```suggestion
    if (value && (value < min || value > max)) {
```

stepchud (approved):
LGTM

PR #294, opened 2024-11-26
[237] Evaluation Form validates uniqueness of Phase

### LINKED ISSUE #237: Evaluation Form validates uniqueness of Phase
# What/Why
A Challenge Phase can have one and only one Evaluation Form associated. This means there should be a unique index on the `evaluation_forms.phase_id`. Users should not be able to select a Phase for a new or existing Form that already has a different Form associated with it.

## Acceptance Criteria
- [x] `evaluation_forms` table has a unique index on `phase_id`
- [x] EvaluationForm model has a Rails `validates :phase_id, :uniqueness`
- [x] The list of challenge phases on the Create/Edit Evaluation Form view should have phases with existing forms filtered out and not available for the user to select in the first place

cpreisinger:
#237


app/views/evaluation_forms/_form.html.erb
```diff
@@ -44,13 +44,15 @@
...
-        <select class="usa-select" required <%= if disabled then "disabled" end %> id="challenge-combo" title="challenge-combo" data-action="evaluation-form#handleChallengeSelect">
-          <% current_user.challenge_manager_challenges.includes(:phases).each do |challenge| %>
-            <% challenge.phases.each do |phase| %>
-              <option value="<%= "#{challenge.id}.#{phase.id}.#{phase.end_date.strftime("%m/%d/%Y")}" %>"><%= challenge_phase_title(challenge, phase) %></option>
-            <% end %>
-          <% end %>
-        </select>
+        <%= select_tag(
+          'challenge-combo',
+          options_for_select(options_for_available_phases(@available_phases)),
```
stepchud, line 49: nice, much cleaner! 🎉

app/views/evaluation_forms/edit.html.erb
```diff
@@ -2,4 +2,4 @@
 
 <h1><%= if eval_form_disabled?(@evaluation_form) then "View" else "Editing" end %> evaluation form</h1>
 
-<%= render "form", evaluation_form: @evaluation_form %>
+<%= render "form", evaluation_form: @evaluation_form, available_phases: @available_phases %>
```
stepchud: you're not using the local variable in the template, but using the instance variable so this isn't necessary to pass in
```suggestion
<%= render "form", evaluation_form: @evaluation_form %>
```

app/views/evaluation_forms/new.html.erb
```diff
@@ -2,4 +2,4 @@
 
 <h1>Create Evaluation Form</h1>
 
-<%= render "form", evaluation_form: @evaluation_form %>
+<%= render "form", evaluation_form: @evaluation_form, available_phases: @available_phases %>
```
stepchud: unused
```suggestion
<%= render "form", evaluation_form: @evaluation_form %>
```

db/migrate/20241125060011_add_unique_phase_constraint_to_evaluation_forms.rb
```diff
@@ -0,0 +1,6 @@
+class AddUniquePhaseConstraintToEvaluationForms < ActiveRecord::Migration[7.2]
+  def change
+    remove_index :evaluation_forms, :phase_id, if_exists: true
+    add_index :evaluation_forms, :phase_id, unique: true
```
stepchud, line 11: I'm unable to run the migration locally because I have some evaluation_forms that have the same `phase_id`. We'll likely have the same problem trying to deploy this to staging too, so it's probably worth fixing now. Easiest thing is just truncate all the evaluation tables to start over from scratch.
```suggestion
    ActiveRecord::Base.connection.truncate_tables(
      EvaluationForm.table_name,
      EvaluationCriterion.table_name,
      Evaluation.table_name,
      EvaluationScore.table_name
    )
    remove_index :evaluation_forms, :phase_id, if_exists: true
    add_index :evaluation_forms, :phase_id, unique: true
```

spec/factories/evaluation_form.rb
```diff
@@ -23,7 +23,11 @@
     # Assures proper points sum of 100 when weighted_scoring = 100
     # Skips initial validation on eval form create because of dependency
     before(:create) do
-      EvaluationForm.skip_callback(:validate, :before, :criteria_weights_must_sum_to_one_hundred)
+      if EvaluationForm._validate_callbacks.any? do |cb|
+        cb.kind == :before && cb.filter == :criteria_weights_must_sum_to_one_hundred
+      end
+        EvaluationForm.skip_callback(:validate, :before, :criteria_weights_must_sum_to_one_hundred)
+      end
```
stepchud, line 30: I'm not sure how this is able to run, but aren't the two `end`s unbalanced?

stepchud (changes requested):
Seems like it's working correctly once I got the migration working, but we'll need to make that change

spec/factories/evaluation_form.rb
stepchud: oh I see, it is a block for the any? filter, the indentation just looks weird

db/migrate/20241125060011_add_unique_phase_constraint_to_evaluation_forms.rb
cpreisinger: Yeah. I figured that would happen just wasn't sure how we wanted to approach it. Truncation is fine

stepchud (approved):

PR #309, opened 2024-12-03
[249] Protect Evaluation Form updates after start_date

### LINKED ISSUE #249: Protect Evaluation Form updates after start_date
# What/Why
In order to ensure stable and accurate evaluations, evaluation forms should not be editable once the evaluation period begins. The start date of evaluation is the same as the end date of the challenge phase submission period. The edit evaluation form on the website disables all fields except the end date of the evaluation period so that challenge managers can always adjust that date. However, this only discourages updates from the frontend and lacks enforcement on the server in case a user submits new data from the form.

The easiest way to prevent this is to discard (not permit) all form params except the end date in the controller. There is another edge case to be aware of. If the challenge manager failed to create an evaluation form by the time the submission period ends, they are allowed to create and save a valid form once, after which the form is disabled for updates.

## Acceptance Criteria
Make the backend changes and add the following specs
- [x] Request spec asserts updates to the form after the start_date are not persisted
  - the response is not required to be error status
- [x] Request spec asserts updates the end_date after the start_date succeed
- [x] Request spec that allows a challenge manager to create one (1) valid form after the start_date, if none exist
  - [x] Assert this form is no longer editable on subsequent requests

## Out of Scope
UI or frontend changes should not be needed

cpreisinger:
#249


spec/requests/evaluation_forms_spec.rb
```diff
@@ -108,6 +108,44 @@
...
 
+  describe "POST /evaluation_forms" do
+    let(:challenge_user) { create_user(role: "challenge_manager") }
+
+    before { log_in_user(challenge_user) }
+
+    context "creating evaluation form after phase end_date" do
+      it "allows creating of a form after phase end_date" do
+        # Generate a challenge and phase that has closed
+        Time.zone.now.to_date
```
stepchud: is this line needed?

spec/requests/evaluation_forms_spec.rb
```diff
@@ -108,6 +108,44 @@
...
+                title: "Example criterion 1",
+                description: "Example description",
+                points_or_weight: 100,
+                scoring_type: :numeric
+              }
+            }
+          }.merge(new_attributes)
+        }
+
+        expect(EvaluationForm.count).to eq(1)
```
stepchud: instead of expecting an exact count before and after, you can expect a change to the value:
```suggestion
        expect {
          post evaluation_forms_path, params: {
            evaluation_form: {
              phase_id: challenge.phases[0].id,
              challenge_id: challenge.id,
              closing_date: 1.day.from_now.to_date,
              evaluation_criteria_attributes: {
                "0" => {
                  title: "Example criterion 1",
                  description: "Example description",
                  points_or_weight: 100,
                  scoring_type: :numeric
                }
              }
            }.merge(new_attributes)
          }
        }.to change(EvaluationForm, :count).by(1)
```

stepchud (approved):
works for me, just left a few comments about the specs

spec/requests/evaluation_forms_spec.rb
cpreisinger: Guess not. Might have accidentally left it in when I was changing some tests around. Or `rubocop -A` did something funky. I'll remove it

cpreisinger:
Might need a rerun from flakey test

Issue #57, opened 2024-07-26
Sort & Filter Submissions


r-bartlett-gsa:
### User Story

As a challenge manager, in order to manage the progress of evaluations efficiently, I would like to be able to sort submissions based on submission and evaluation progress data.

**Acceptance criteria:**

A challenge manager can sort and filter the submissions (on the submission list view) by:
- [x]  Filter by: Progress status: completed, in progress, not started, recused
- [x]  Filter by: Eligible for Evaluations (Prescreen status)
- [x]  Filter by: Advancement status
- [x] Sort by: Submission ID: low to high, high to low
- [x] #418
- [x] Sort by: Average score (for completed evaluations only): low to high, high to low

**Dev tasks:**

- [x] #207

### Definition of Done

**Doing (dev team)**
- [x] Code complete 
- [x] Code is organized appropriately
- [x] Any known trade offs are documented in the associated GH issue
- [x] Code is documented, modules, shared functions, etc.
- [x] Automated testing has been added or updated in response to changes in this PR
- [x] The feature is smoke tested to confirm it meets requirements
- [x] Database changes have been peer reviewed for index changes and performance bottlenecks
- [x] PR that changes or adds UI
  - [x] include a screenshot of the WAVE report for the altered pages
  - [x] Confirm changes were validated for mobile responsiveness
- [x] PR approved / Peer reviewed 
- [x] Security scans passed
- [x] Automate accessibility tests passed
- [x] Build process and deployment is automated and repeatable
- [x] Feature toggles if appropriate
- [x] Deploy to staging
- [x] Move card to testing column in the board
---
**Staging**
- [ ] Accessibility tested (Marni)
  - [ ]  Keyboard navigation
  - [ ]  Focus confirmed
  - [x] Color contrast compliance
  - [ ] Screen reader testing 
- [x] Usability testing: mobile and desktop (Tracy or Marni)
- [x] Cross browser testing - UI rendering is performant on below listed devices/browsers (Tracy or Marni)
  - [x] Windows/Chrome
  - [x] Windows/Edge
  - [x] Mac/Chrome
  - [x] Mac/Safari
  - [x] iOS/Safari
- [x] AC review (Renata)
- [x] Deploy to production (production-like environment for eval capability) (dev team)
- [x] Move to production column in the board
---
**Production**
- [ ] User and security documentation has been reviewed for necessary updates (Renata/Tracy/Dev team) 
- [ ] PO / PM approved (Jarah or Renata)
[...]

PR #313, opened 2024-12-05
[207] Sort and Filter Component

### LINKED ISSUE #207: Create the Sort & Filter component
# What/Why
This feature is important for challenges with many submissions so the CM can navigate them more efficiently and find what they need. Sorts and filters are applied on the backend as we will not load all submissions on the page at once.

Make sure you read the parent ticket including the comment thread.

### Design
Screenshot 2024-10-10 at 3 39 13 PM

## Acceptance Criteria
- [x] Implements the Sort & Filter dropdown and updates the list of submissions appropriately.
  - [x] The first 6 entries are filters that (possibly) reduce the number of submissions on the page
  - [x] The last 4 are sort filters that re-order the results
- [x] This should probably be a param in the URL so that history is maintained and reloads the page for each change
- [x] Automated tests for the behaviors

- [x] WAVE screenshot

emmabjj:
Related tickets: #57 & #207 

As a challenge manager, in order to manage the progress of evaluations efficiently, I would like to be able to sort submissions based on submission and evaluation progress data.
- Filter
  - Statuses:
    - completed 
    - in progress
    - not started
    - recused
  - Eligible for Evaluations
  - Selected to advance
- Sorting
  - Submission ID: low to high & high to low
  - Average score: low to high & high to low
  
sort_and_filter

Screenshot 2024-12-05 at 5 29 34 PM

Screenshot 2024-12-05 at 5 31 09 PM

Updated UI:
Screenshot 2024-12-27 at 1 58 34 PM
Screenshot 2024-12-27 at 1 58 59 PM


app/controllers/phases_controller.rb
```diff
@@ -11,6 +11,23 @@ def index
 
   def submissions
     @submissions = @phase.submissions
+
+    @not_started = @submissions.where.missing(:evaluations)
```
stepchud: til! never seen the `missing` scope before cool!

app/controllers/phases_controller.rb
```diff
@@ -19,4 +36,42 @@ def set_phase
...
+  end
+
+  def filter_recused_submissions
+    @submissions.joins(:evaluator_submission_assignments).
+      where(evaluator_submission_assignments: { status: :recused })
+  end
+
+  def apply_eligibility_filters
+    @submissions = @submissions.select(&:eligible_for_evaluation?) if params[:eligible_for_evaluation] == 'true'
+    @submissions = @submissions.select(&:selected_to_advance?) if params[:selected_to_advance] == 'true'
```
stepchud: for performance it might be better to apply this filter to the SQL query rather than in-memory. actually these could be useful scopes on the model.
```suggestion
    @submissions = @submissions.where(judging_status: [:selected, :winner]) if params[:eligible_for_evaluation] == 'true'
    @submissions = @submissions.where(judging_status: :selected) if params[:selected_to_advance] == 'true'
```

app/models/submission.rb
```diff
@@ -67,4 +68,23 @@ def eligible_for_evaluation?
...
+    direction_sql = direction == :desc ? 'DESC' : 'ASC'
+
+    joins("LEFT JOIN evaluations ON evaluations.submission_id = submissions.id").
+      group('submissions.id').
+      order(
+        Arel.sql(
+          "COALESCE(ROUND(AVG(evaluations.total_score)), 0) #{direction_sql}, " \
+          "submissions.id #{direction_sql}"
+        )
+      )
```
stepchud: pretty cool, eventually I think we might need to exclude incomplete evaluations

app/views/phases/_sort_and_filter.html.erb
```diff
@@ -0,0 +1,59 @@
...
+            >
+              <span style="color: white !important;">Sort & Filter</span>
+            </button>
+            <style>
+              .usa-accordion__button.usa-current::after {
+                content: none !important;
+              }
+              .usa-nav__primary button[aria-expanded="false"] span::after {
+                background-color: white !important;
+              }
```
stepchud: this `<style>` tag seems pretty out of place here , can you describe what it's doing? might be better to extract this to a normal app css/sass with a custom class to target the component instead of the generic USWDS classes.

app/views/phases/_sort_and_filter.html.erb
```diff
@@ -0,0 +1,59 @@
...
+        <ul class="usa-nav__primary usa-accordion">
+          <li class="bg-primary radius-md">
+            <button
+              type="button"
+              class="usa-accordion__button usa-nav__link usa-current"
+              aria-expanded="false"
+              aria-controls="basic-nav-section-one"
+              style="&:hover { color: inherit !important; outline: none; }"
+            >
+              <span style="color: white !important;">Sort & Filter</span>
```
stepchud: do these style attributes need the !important rule? i think we can extract these custom styles to css/sass too

app/views/phases/_submissions_table.html.erb
```diff
@@ -6,34 +6,37 @@
...
         <th scope="col">Average Score</th>
+        <th scope="col"><span class="usa-sr-only">View Submission</span></th>
       </tr>
     </thead>
     <tbody>
       <% @submissions.each do |submission| %>
           <tr>
             <th data-label="Submission ID" scope="row">
-              <span class="mobile-lg:display-none">Submission ID </span><%= submission.id %>
+              <span class="mobile-lg:display-none">Submission ID</span><%= submission.id %>
```
stepchud: Screenshot 2024-12-10 at 4 16 16 PM

The mobile layout is missing the space between the label and id, should revert this
```suggestion
              <span class="mobile-lg:display-none">Submission ID </span><%= submission.id %>
```

app/javascript/controllers/sort_filter_menu_controller.js
stepchud, line 1: nice controller 👏

app/views/phases/_sort_and_filter.html.erb
```diff
@@ -0,0 +1,59 @@
+<div class="grid-container padding-x-0">
+  <div class="grid-row">
+    <div class="grid-col-12 display-flex flex-column flex-align-start padding-top-105">
+      <nav aria-label="Sort and filter options" class="usa-nav" data-controller="sort-filter-menu">
```
stepchud: the nav is hidden below 1024 because of a style on the `usa-nav` class but the figma designs show this dropdown available on mobile.

spec/requests/phases_spec.rb
```diff
@@ -81,4 +81,111 @@
...
+    let(:challenge) { create(:challenge, user: challenge_user) }
+    let(:phase) { create(:phase, challenge: challenge) }
+    let!(:submission1) { create(:submission, challenge: challenge, phase: phase, status: 'submitted') }
+    let!(:submission2) { create(:submission, challenge: challenge, phase: phase, status: 'submitted') }
+    let!(:submission3) { create(:submission, challenge: challenge, phase: phase, status: 'submitted', judging_status: 'selected') }
+    let!(:submission4) { create(:submission, challenge: challenge, phase: phase, status: 'submitted', judging_status: 'winner') }
+
+    before { log_in_user(challenge_user) }
+
+    it 'assigns submissions and calculates status counts' do
```
stepchud: there is a principle in testing that you test the boundaries of the object and treat the internals as a black box. this spec tests the implementation details of the controller by asserting which instance variables are assigned. with higher level specs like a request spec, you really want to assert that the response body (output) is correct based on the request data (input), not the internals of how the server gets there while processing the request.

I would probably replace this spec with one that asserts the `submissions_by_status` counts are displayed to the user in the response body. if you wanted you could assert that all submissions are visible (none are filtered).

stepchud (changes requested):
Left some feedback

app/controllers/phases_controller.rb
```diff
@@ -11,6 +11,23 @@ def index
...
   def submissions
     @submissions = @phase.submissions
+
+    @not_started = @submissions.where.missing(:evaluations)
+
+    @in_progress = @submissions.joins(:evaluations).
+      where(evaluations: { completed_at: nil })
+
+    @completed = @submissions.joins(:evaluations).
+      where.not(evaluations: { completed_at: nil })
```
stepchud: check for uniqueness on joins(:evaluations)

spec/requests/submissions_spec.rb
```diff
@@ -111,4 +111,129 @@
...
+
+      context 'when filtering submissions' do
+        it 'shows only submissions matching the selected status' do
+          get submissions_phase_path(phase), params: { status: 'not_started' }
+
+          expect(response.body).to include(not_started_submission.id.to_s)
+          expect(response.body).to include(eligible_submission.id.to_s)
+          expect(response.body).not_to include(selected_submission.id.to_s)
+          expect(response.body).not_to include(in_progress_submission.id.to_s)
+          expect(response.body).not_to include(completed_submission.id.to_s)
```
stepchud: you have some specs that are failing inconsistently related to these expects. when you expect that an ID is not anywhere in the response body, it makes it likely that some number like 10 is present somewhere in the response from the server failing the spec even though it's not actually present as an ID. I think what you want to do here is add some more specificity to your query instead of being very broad, for example by using the `have_css` selector.
if it helps you can also add `data-` attributes to your html markup to make it easier to target those specific elements in your tests.

emmabjj: Ahhh I see, thank you @stepchud

spec/requests/submissions_spec.rb
```diff
@@ -111,4 +111,129 @@
...
+        end
+
+        it 'orders submissions by score high to low' do
+          get submissions_phase_path(phase), params: { sort: 'average_score_high_to_low' }
+
+          response_body = response.body
+          high_score_index = response_body.index(completed_submission.id.to_s)
+          low_score_index = response_body.index(in_progress_submission.id.to_s)
+
+          expect(high_score_index).to be < low_score_index
```
stepchud: this spec is also failing I believe because of the same broadness in finding content on the page. we have [Capybara::RSpecMatchers](https://rubydoc.info/github/jnicklas/capybara/Capybara/RSpecMatchers) available for use in request specs. take a look at those and see if something might be more helpful.

emmabjj:
@stepchud - thanks for help! Those tests weren't failing locally so it was difficult to reproduce, but looks like the tests are good now.

spec/requests/submissions_spec.rb
```diff
@@ -217,21 +217,19 @@
...
           get submissions_phase_path(phase), params: { sort: 'average_score_high_to_low' }
 
-          response_body = response.body
-          high_score_index = response_body.index(completed_submission.id.to_s)
-          low_score_index = response_body.index(in_progress_submission.id.to_s)
-
-          expect(high_score_index).to be < low_score_index
+          expect(response.body).to have_selector(
+            "tr[data-submission-id='#{completed_submission.id}']" \
+            " ~ tr[data-submission-id='#{in_progress_submission.id}']"
```
stepchud, line 222: 🎉  not sure I've ever used the `~` selector, so that's cool 😀

stepchud (changes requested):
Some UI styling issues that I discovered. I wonder if the custom css in application.sass.scss needs to be tweaked
1. Mobile/tablet layout the menu hover color is low contrast
2. The border of the nav menu items is transparent and shows the text behind it

Screenshot 2024-12-20 at 12 26 23 AM

stepchud:
another small UI issue:
3. The dropdown style should be consistent, on smaller layout is +/- instead of the ^ / ⌄
Screenshot 2024-12-20 at 9 24 12 AM
Screenshot 2024-12-20 at 9 24 19 AM


> Some UI styling issues that I discovered. I wonder if the custom css in application.sass.scss needs to be tweaked
> 
> 1. Mobile/tablet layout the menu hover color is low contrast
> 2. The border of the nav menu items is transparent and shows the text behind it
> 
> Screenshot 2024-12-20 at 12 26 23 AM

emmabjj:
> another small UI issue: 3. The dropdown style should be consistent, on smaller layout is +/- instead of the ^ / ⌄ Screenshot 2024-12-20 at 9 24 12 AM Screenshot 2024-12-20 at 9 24 19 AM
> 
> > Some UI styling issues that I discovered. I wonder if the custom css in application.sass.scss needs to be tweaked
> > 
> > 1. Mobile/tablet layout the menu hover color is low contrast
> > 2. The border of the nav menu items is transparent and shows the text behind it
> > 
> > Screenshot 2024-12-20 at 12 26 23 AM

I'll fix that contrast error - lots of little tweaking yesterday to make it responsive and keep the same styling between different screen sizes with USWDS. There was some real weirdness with that built-in USWDS +/- dropdown icon for mobile, I was having a difficult time overriding the styling. I'll give it another go, I wanted to push the other fixes first. Thanks @stepchud !

app/models/submission.rb
```diff
@@ -36,7 +36,7 @@ class Submission < ApplicationRecord
   belongs_to :manager, class_name: 'User'
   has_many :evaluator_submission_assignments, dependent: :destroy
   has_many :evaluators, through: :evaluator_submission_assignments, class_name: "User"
-  has_many :evaluations, dependent: :destroy
+  has_many :evaluations, through: :evaluator_submission_assignments, dependent: :destroy
```
stepchud: the `evaluations` table has the `submission_id` with an index on it (Evaluation model `belongs_to :submission`). we can use that more efficiently to query directly instead of going through another table.
```suggestion
  has_many :evaluations, dependent: :destroy
```

app/models/submission.rb
```diff
@@ -70,4 +70,26 @@ def eligible_for_evaluation?
   def selected_to_advance?
     winner?
   end
+
+  def average_score
+    avg = evaluations.average(:total_score)
+    score = avg ? avg.round : 0
+    [score, score.to_s]
```
stepchud: it's unusual to me that the `Submission#average_score` method returns a two element array, I would expect it to return a value. seems unnecessary since you can always call .to_s on the result of the calculation when needed. I'd probably refactor this to simplify so it only calculates the average especially since it's only used one place. looking at that call site, I don't think it's necessary to convert the number to a string anyways. the interpolation `<%= score %>` should convert that int/float value to a string in the submissions_table html template for you already (tested locally, seems to work fine).

app/controllers/phases_controller.rb
```diff
@@ -26,4 +25,80 @@ def set_phase
...
+
+  def set_submission_statuses
+    @not_started = @submissions.where.missing(:evaluations)
+    @in_progress = @submissions.joins(:evaluations).
+      where(evaluations: { completed_at: nil }).distinct
+    @completed = @submissions.joins(:evaluations).
+      where.not(evaluations: { completed_at: nil }).distinct
+
+    @submissions_by_status = {
+      not_started: @not_started.count,
```
stepchud, line 44: ~I'm seeing these stats locally which doesn't make sense to me. how can I have 5 not started when there is only 1 submission?~
Screenshot 2025-01-02 at 1 20 43 PM
EDIT: nevermind, I have some funky duplicated evaluator_submission_assignments records locally, but FWIW this is fixed by removing the :through association from submission has_many :evaluations as suggested on the model

app/controllers/phases_controller.rb
```diff
@@ -26,4 +25,80 @@ def set_phase
...
+    @eligible_count = @submissions.eligible_for_evaluation.count
+    @selected_count = @submissions.winner.count
+  end
+
+  def set_submission_statuses
+    @not_started = @submissions.where.missing(:evaluations)
+    @in_progress = @submissions.joins(:evaluations).
+      where(evaluations: { completed_at: nil }).distinct
+    @completed = @submissions.joins(:evaluations).
+      where.not(evaluations: { completed_at: nil }).distinct
```
stepchud: Screenshot 2025-01-02 at 6 40 16 PM
There is an issue in the completed query (shown in the incorrect status counts that don't add up to the total count) where it is including submissions where *any* evaluations are completed. the correct behavior should be showing submissions where *all* evaluations are completed. it took me a while to figure out, but I think this does the trick the simplest way possible:
```suggestion
    @completed = @submissions.joins(:evaluations).
      where.not(evaluations: @in_progress).distinct
```
translation: select the submissions that have at least one evaluation where none of the evaluations are in the @in_progress subquery.

stepchud (changes requested):
Super close, just a few more backend issues.
Also I'm still seeing the issue "Mobile/tablet layout the menu hover color is low contrast" from my previous comment but I'm not sure that's a blocker since mobile/tablet users can't really hover on a touchscreen. It's just weird that it shows up at that breakpoint, I guess there is something built into the uswds css. I tracked it down to this specific rule if you want to see checkout a way to override that:
Screenshot 2025-01-02 at 7 52 23 PM

Screenshot 2025-01-02 at 7 41 49 PM

app/controllers/phases_controller.rb
emmabjj: Okay interesting, thanks! I had some submissions that are recused but still assigned, so they factored into the total submissions count but not the completed/in progress/not started counts, I thought it was coming from that.

app/models/submission.rb
emmabjj: This was part of the design for formatting the scoring (94 & 94%) that I was following in the figma, but now the score is no longer required to be a percentage, this code is out of date.

stepchud (approved):
🔥 Works for me there are some specs failing though

PR #317, opened 2024-12-10
[30] Proxy Static Pages

### LINKED ISSUE #30: Proxy Challenge.gov Pages through Rails app
- [x] Set up the Rails application to load static pages hosted by the Pages build so that those pages remain accessible on Challenge.gov/resources etc
- [x] Proxy all URLs that are not / and /challenges or /challenge (this should be resources, user-guide, blog, terms, about, contact, privacy-policy)

danivovich:
Map a proxy such that URLs attempt to load jekyll content from the static pages content.

Implements #30


config/routes.rb
```diff
@@ -35,13 +34,19 @@
...
   if Rails.env.development? || Rails.env.dev? || Rails.env.test?
     namespace :dev do
       get "/sandbox", to: "sandbox#index"
       get "/accounts", to: "accounts#index"
       post "/login", to: "accounts#login"
     end
   end
+
+  if Rails.env.test?
+    get '/', to: "dashboard#index"
```
stepchud: do we actually want this? it seems okay if we just get rid of the root path routing to the dashboard index and either remove the tests that assert it loads the dashboard or update them to assert redirected to `/dashboard`

app/controllers/pages_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+        end
+      end
+    end
+  end
+
+  private
+
+  def rewrite_links(html)
+    parsed_html = html.gsub("https://challenge.gov/", "/")
+    parsed_html = parsed_html.gsub(HOST, "/")
```
stepchud: I wonder if we need to check if this is an href attribute or something else e.g. plain text in the page content.

config/routes.rb
stepchud: not a blocker, I just noticed the root path routing so it's a little funky, but I'm not sure it's better to have the test env with such a difference from other envs. I'd probably just remove any specs that test the root path, or possibly do something like mock the proxied page? 🤔

stepchud:
one smaller issue is that the login and sign up button in the header go to the production login.gov site, so devs will need to log in to localhost via the `/dashboard` route (any 404 redirects there)

app/controllers/pages_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+          response.body = rewrite_links(response.body)
+        end
+      end
+    end
+  end
+
+  private
+
+  def rewrite_links(html)
+    parsed_html = html.gsub("https://challenge.gov/", "/")
```
stepchud: just noting the commented-out DOMAIN and HOST use `www.challenge.gov`, is the distinction between www subdomain significant somehow?

stepchud (commented):
maybe this is a dumb question or I don't understand how it works, but if the rails app is hosted on `www.challenge.gov`, and we reverse proxy to that same domain, won't it just create an infinite loop? are we migrating the jekyll server to a different domain?

app/controllers/pages_controller.rb
danivovich: Yeah I think this is a little bit stopgap, once the static content is only served this way, we can adjust how it sets it urls. Also it is feasible to have assets go to the static page domain, and not proxy everything, just the html.

app/controllers/pages_controller.rb
danivovich: yeah this may have been holdover. adjusted.

config/routes.rb
danivovich: I un-messed this

spec/requests/pages_spec.rb
```diff
@@ -0,0 +1,43 @@
+require "rails_helper"
+
+RSpec.describe "PagesController" do
+  it "get new renders successfully" do
+    stub_request(:get, PagesController::HOST + PagesController::BASE_URL).
```
stepchud: 👍

stepchud (changes requested):
The links to the active challenges on the homepage are broken. The URL path it is constructing is basically doubled up (see `path=` in the log output). Seems like the gsub for rewriting the links on the page isn't working on the BASE_URL gsub correctly because the links include that in the path still. Another issue is that it's not including the params in the request but I think the jekyll server needs the challenge param, e.g. `?challenge=eval-challenge-test`, to fetch the correct challenge.
Screenshot 2024-12-13 at 2 01 34 PM

Here are some request logs where I put in some debug logging:
```
13:55:17 web.1   | Started GET "/preview/gsa/challenges-and-prizes/eval-dev/?challenge=eval-challenge-test" for ::1 at 2024-12-13 13:55:17 -0800
13:55:17 web.1   | Processing by PagesController#index as HTML
13:55:17 web.1   |   Parameters: {"challenge"=>"eval-challenge-test", "path"=>"preview/gsa/challenges-and-prizes/eval-dev"}
13:55:18 web.1   | path=/preview/gsa/challenges-and-prizes/eval-dev/preview/gsa/challenges-and-prizes/eval-dev/
13:55:18 web.1   | Redirected to http://localhost:3000/dashboard
13:55:18 web.1   | Completed 302 Found in 352ms (ActiveRecord: 0.0ms (0 queries, 0 cached) | GC: 0.2ms)
```

stepchud:
> The links to the active challenges on the homepage are broken. The URL path it is constructing is basically doubled up (see `path=` in the log output). Seems like the gsub for rewriting the links on the page isn't working on the BASE_URL gsub correctly because the links include that in the path still. Another issue is that it's not including the params in the request but I think the jekyll server needs the challenge param, e.g. `?challenge=eval-challenge-test`, to fetch the correct challenge. Screenshot 2024-12-13 at 2 01 34 PM
> 
> Here are some request logs where I put in some debug logging:
> 
> ```
> 13:55:17 web.1   | Started GET "/preview/gsa/challenges-and-prizes/eval-dev/?challenge=eval-challenge-test" for ::1 at 2024-12-13 13:55:17 -0800
> 13:55:17 web.1   | Processing by PagesController#index as HTML
> 13:55:17 web.1   |   Parameters: {"challenge"=>"eval-challenge-test", "path"=>"preview/gsa/challenges-and-prizes/eval-dev"}
> 13:55:18 web.1   | path=/preview/gsa/challenges-and-prizes/eval-dev/preview/gsa/challenges-and-prizes/eval-dev/
> 13:55:18 web.1   | Redirected to http://localhost:3000/dashboard
> 13:55:18 web.1   | Completed 302 Found in 352ms (ActiveRecord: 0.0ms (0 queries, 0 cached) | GC: 0.2ms)
> ```

FYI, this URL is working locally: http://localhost:3000/?challenge=eval-challenge-test

stepchud:
oh it's because the react app is loading those from the data API after the initial page load 🫤

stepchud (approved):
LGTM

PR #321, opened 2024-12-13
[291 & 292] Evaluation Status Details

### LINKED ISSUE #291: Add evaluation status details to the list
# What/Why
As a challenge manager I want to view the details of evaluators and their evaluation progress for each submission.
As a challenge manager I want to quickly see a scoring average for each submission in my challenge.

### Figma Link

[Submission index list](https://www.figma.com/design/VfdBFG8ShVnOStdSNYy9fR/Challenge.Gov-Component-Library?node-id=2974-59388&t=KdJ650lG6P1BP9OR-1)

## Technical Notes
Use [tag](https://designsystem.digital.gov/components/tag/) with colors for progress status if convenient, otherwise make a new reusable component for the app.
As a reference, some progress queries can be seen on #280 

## Acceptance Criteria
- [x] Name of the evaluator is listed
  - [x] Evaluator name is highlighted in red if they have recused
- [x] Status tags are colored according to Figma design
- [x] Average score is shown when all evaluators have competed their scoring
  - [x] otherwise the score is displayed `N/A`
- [x] Mobile layout is responsive

- [x] WAVE screenshot

### LINKED ISSUE #292: Update submission Eligible for Evaluation / Selected to Advance from list view
# What/Why
As a challenge manager, I want to mark submissions as eligible for evaluation from the list view for my ease of use.
As a challenge manager, I want to mark submissions as selected to advance from the list view for convenience.

### Figma Design
Screenshot 2024-11-25 at 5 11 49 PM
Screenshot 2024-11-25 at 5 12 25 PM

## Acceptance Criteria
- [x] Both buttons update the `judging_status` attribute on the Submission record
  - [x] route - `PATCH /submissions/:id`
- [x] access is limited to the challenge managers of the phase of the submission
- [x] Desktop view shows the options as a checkbox
- [x] Mobile view shows the options as [unstyled buttons](https://designsystem.digital.gov/components/button/) (`class="usa-button usa-button--unstyled"`)
- [x] _Selected to advance_ is disabled (deselected and can't be selected) when the submission is not _Eligible for evaluation_ status
- [x] _Eligible for evaluation_ status is disabled (selected can't be deselected) when the submission has evaluators assigned to it
  - [from this comment](https://github.com/GSA/Challenge_platform/issues/292#issuecomment-2504741408)
  - Evaluators will only be assignable _after_ the submission is marked _Eligible for evaluation_, but that requirement is not yet implemented and out of scope here

- [x] WAVE screenshot

## Out of Scope
It should not be possible to assign evaluators to a submission that is not marked eligible for evaluation. However this requirement is out of scope for this ticket and will not be implemented/enforced yet.

emmabjj:
Related tickets: #291 #292 

Add evaluation status details in the submission index for a challenge phase.

Details added in this PR:
- **Evaluator Name**
  - Evaluator name is highlighted in red if recused
- **Evaluation Status** tag for the evaluator's progress on the evaluation for that submission (recused, completed, in progress, not started)
- **Average score** for the evaluations of that submission
- **Judging status checkbox functionality** (_eligible for evaluation & selected to advance_)
  - These checkboxes are disabled under these conditions
    - Eligible for evaluation is disabled when:
      - the submission has evaluators assigned to it
    - Selected to advance is disabled when:
      - not eligible for evaluation
      - when there are evaluations assigned to the submission that are incomplete
      - when there are no evaluators assigned to the submission.

Screenshot 2024-12-13 at 10 20 16 AM
Screenshot 2024-12-13 at 10 20 29 AM
Screenshot 2024-12-13 at 10 21 02 AM


app/controllers/phases_controller.rb
```diff
@@ -26,4 +29,15 @@ def set_phase
     @phase = Phase.where(challenge: current_user.challenge_manager_challenges).find(params[:id])
     @challenge = @phase.challenge
   end
+
+  def include_evaluator_associations
+    return unless request.format.html? && @submissions.any?
+    return unless evaluator_assignments?
```
stepchud: is this check necessary or can we always include the associations? I think the submissions will usually have these assignments while the evaluation window is open (and the CM is checking progress).

app/controllers/submissions_controller.rb
```diff
@@ -8,6 +8,28 @@ class SubmissionsController < ApplicationController
   def show; end
 
   def update
+    if params[:submission][:judging_status].present?
+      handle_judging_status_update
+    else
+      handle_comments_update
+    end
```
stepchud: as Stone mentioned we will need to handle an update to both judging_status and comments in a single request. it should still return an error if the judging_status is an invalid option. probably the best way is to move the validation to the Submission model so that rails handles the validation in a standard way.

stepchud (commented):
still reviewing but I wanted to send the feedback on the `update` route so you have it sooner

app/controllers/submissions_controller.rb
```diff
@@ -8,6 +8,28 @@ class SubmissionsController < ApplicationController
...
+  end
+
+  private
+
+  def handle_judging_status_update
+    unless valid_status_change?
+      return respond_with_error
+    end
+
+    if @submission.update(submission_params)
```
stepchud: as an example of how to write the model validation, you can add the method like this... [more details here](https://guides.rubyonrails.org/active_record_validations.html#performing-custom-validations)
```
class Submission < ApplicationRecord
  validate :can_be_selected_to_advance, if: &:winner? 
  validate :can_be_ineligible_for_evaluation if: -> { judging_status_change == [:selected, :not_selected] }

  private
  def can_be_selected_to_advance
    errors.add(:judging_status, "can't be selected to advance when not eligible for evaluation") unless judging_status_was == :selected
    errors.add(:judging_status, "can't be selected to advance until all evaluations are complete") unless all_evaluations_completed?
  end
end
```

```suggestion
    if @submission.update(submission_params)
```

spec/requests/submissions_spec.rb
```diff
@@ -58,6 +64,11 @@
         create(:submission, challenge: challenge, phase: phase)
         create(:submission, challenge: challenge, phase: phase, judging_status: "selected")
 
+        allow_any_instance_of(ActionView::Base).to receive(:render).and_call_original
+        allow_any_instance_of(ActionView::Base).to receive(:render).
+          with(hash_including(partial: "submissions_table")).
+          and_return("")
```
stepchud: why do we need to mock the views here?

spec/requests/submissions_spec.rb
```diff
@@ -111,4 +122,103 @@
...
+                 submission: submission,
+                 evaluator: evaluator,
+                 status: :assigned)
+
+          patch submission_path(submission), params: {
+            submission: { judging_status: 'selected' }
+          }, as: :json
+
+          expect(response).to have_http_status(:unprocessable_entity)
+          expect(submission.reload.judging_status).to eq('not_selected')
```
stepchud: this spec is confusing, I thought we block the update to not_selected when there are assignments? but this tests that we block a judging_status update to selected, which I think is an invalid state (it can't be not_selected if there are assignments)

stepchud: also the title in incorrect since we don't prevent all eligibility changes, only the change from select => not_selected if there are evaluator assignments.

spec/requests/submissions_spec.rb
```diff
@@ -111,4 +122,103 @@
...
+        end
+
+        it "prevents changing judging_status from 'not_selected' to 'selected' when eligibility checkbox is disabled" do
```
stepchud: can we reverse the direction here and test that it
```suggestion
        it "prevents changing judging_status from 'selected' to 'not_selected' when eligibility checkbox is disabled" do
```

stepchud: (i think this and the previous spec are testing the same thing basically?)

app/models/submission.rb
```diff
@@ -70,4 +70,20 @@ def eligible_for_evaluation?
   def selected_to_advance?
     winner?
   end
+
+  def eligibility_checkbox_disabled?
+    evaluators.any?
+  end
```
stepchud: I tried to unassign an evaluator from the submission but this query still returns true and the checkbox is not re-enabled. it should filter for the evaluators where assignment status in ['assigned', 'recused'] so we don't flag 'unassigned' or 'recused_unassigned' as being evaluators.
```suggestion
  def eligibility_checkbox_disabled?
    evaluators.where(status: [:assigned, :recused]).exists?
  end
```

app/controllers/phases_controller.rb
stepchud: we'll leave this in

emmabjj:
@stepchud I keep getting the unoptimized query error:

```
     Bullet::Notification::UnoptimizedQueryError:
       user: circleci
       GET /phases/248/submissions
       AVOID eager loading detected
         Submission => [:evaluators]
         Remove from your query: .includes([:evaluators])
       Call stack
         /home/circleci/Challenge_platform/app/views/phases/_submissions_table.html.erb:20:in `_app_views_phases__submissions_table_html_erb___3569867229421565334_71700'
         /home/circleci/Challenge_platform/app/views/phases/submissions.html.erb:10:in `_app_views_phases_submissions_html_erb___2327403691024650102_67960'
         /home/circleci/Challenge_platform/spec/requests/submissions_spec.rb:50:in `block (4 levels) in <top (required)>'
```

The error bounces between USE and AVOID when I change it.

app/controllers/submissions_controller.rb
```diff
@@ -8,64 +8,16 @@ class SubmissionsController < ApplicationController
...
-    end
-
-    true
-  end
-
-  def respond_with_success
     respond_to do |format|
-      format.html { redirect_to submissions_phase_path(@submission.phase) }
-      format.json { render json: { success: true }, status: :ok }
+      if @submission.update(submission_params)
```
stepchud, line 12: 🍰

app/models/submission.rb
```diff
@@ -86,4 +90,21 @@ def all_evaluations_completed?
...
+
+  def can_be_selected_to_advance
+    unless judging_status_was == 'selected'
+      errors.add(:judging_status, "can't be selected to advance when not eligible for evaluation")
+      return
+    end
+
+    return unless advancement_checkbox_disabled?
+
+    errors.add(:judging_status, "can't be selected to advance if not all evaluations are complete")
```
stepchud: nit: copy
```suggestion
    errors.add(:judging_status, "can't be selected to advance until all evaluations are complete")
```

app/models/submission.rb
```diff
@@ -86,4 +90,21 @@ def all_evaluations_completed?
       includes(:evaluation).
       all? { |assignment| assignment.evaluation_status == :completed }
   end
+
+  def can_be_selected_to_advance
+    unless judging_status_was == 'selected'
```
stepchud: this shouldn't be possible since the `validate if: judging_status_change` already checks the previous value was `selected`, otherwise this validation won't be triggered

app/models/submission.rb
```diff
@@ -86,4 +90,21 @@ def all_evaluations_completed?
...
+
+    return unless advancement_checkbox_disabled?
+
+    errors.add(:judging_status, "can't be selected to advance if not all evaluations are complete")
+  end
+
+  def can_be_ineligible_for_evaluation
+    return unless eligibility_deselection_disabled?
+
+    errors.add(:judging_status, "can't deselect evaluation eligibility when there are evaluators assigned")
```
stepchud: nit: copy
```suggestion
    errors.add(:judging_status, "must remain eligible for evaluation when evaluators are assigned")
```

app/models/submission.rb
```diff
@@ -71,8 +75,8 @@ def selected_to_advance?
     winner?
   end
 
-  def eligibility_checkbox_disabled?
-    evaluators.any?
+  def eligibility_deselection_disabled?
+    evaluator_submission_assignments.exists?(status: [:assigned, :recused])
   end
 
   def advancement_checkbox_disabled?
```
stepchud: for model method names I prefer to avoid references to UI state (e.g. disabled) and refer to the state of model instead such as:
```suggestion
  def evaluators_assigned?
    evaluator_submission_assignments.exists?(status: [:assigned, :recused])
  end

  def evaluations_missing_or_incomplete?
```

spec/requests/submissions_spec.rb
stepchud, line 1: nice specs! 🙌

app/models/submission.rb
stepchud: I think you can fix the Bullet Unoptimized query "USE eager loading detected" by [changing this line](https://github.com/GSA/Challenge_platform/blob/bde972de60cc063c08db949de8f440b6f0d3e2b3/app/models/submission.rb#L83):
https://github.com/GSA/Challenge_platform/blob/bde972de60cc063c08db949de8f440b6f0d3e2b3/app/models/submission.rb#L83
by removing the reference to `evaluations` and using the assignment assocation instead:
```ruby
!eligible_for_evaluation? || !all_evaluations_completed? || evaluator_submission_assignments.empty?
```

app/controllers/submissions_controller.rb
```diff
@@ -75,4 +27,17 @@ def submission_params
   def set_submission
     @submission = Submission.by_user(current_user).find(params[:id])
   end
+
+  def handle_successful_update(format)
+    format.html do
+      flash[:success] = I18n.t("submission_updated")
+      redirect_to submission_path(@submission.phase)
```
stepchud: this should redirect to the submission id, not the phase id
```suggestion
      redirect_to submission_path(@submission)
```

stepchud (changes requested):
just a couple smaller updates otherwise this looks good 👍

app/models/submission.rb
```diff
@@ -71,6 +75,33 @@ def selected_to_advance?
...
+    return unless evaluations_missing_or_incomplete?
+
+    errors.add(:judging_status, "can't be selected to advance until all evaluations are complete")
+  end
+
+  def can_be_ineligible_for_evaluation
+    return unless evaluators_assigned?
+
+    errors.add(:judging_status, "must remain eligible for evaluation when evaluators are assigned")
+
```
stepchud:
```suggestion
  end

```

app/helpers/evaluations_helper.rb
```diff
@@ -0,0 +1,29 @@
...
+    completed_evaluations = submission.evaluations.where.not(completed_at: nil)
+
+    unless completed_evaluations.any?
+      return Score.new(0, "0", "N/A")
+    end
+
+    avg = completed_evaluations.average(:total_score)
+    score = avg ? avg.round : 0
+    Score.new(score, score.to_s, score.to_s)
+  end
```
stepchud: I think these helpers are all unused now and can be removed?

emmabjj: Sorry, I missed this comment. These are being used in the submissions index for challenge managers and evaluators, the formatting for average scoring is different between the two. I've combined a few different scoring methods I've found, so I'll push that change to this PR.

emmabjj:
@stepchud Sorting by average score and submission id is no longer working

app/controllers/phases_controller.rb
```diff
@@ -10,30 +10,25 @@ def index
...
-
-    include_evaluator_associations
+    @submissions = @phase.submissions.includes(evaluator_submission_assignments: [:evaluator, :evaluation])
 
     set_submission_counts
     set_submission_statuses
 
-    apply_filters
-    apply_sorting
+    @submissions = SortAndFilterService.new(
```
stepchud: 🎉

app/services/sort_and_filter_service.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+# This service handles sort and filtering submissions.
+class SortAndFilterService
```
stepchud: nit: could be named more specific e.g.
```suggestion
class SubmissionsSortAndFilterService
```

app/views/phases/_submissions_table_rows.html.erb
```diff
@@ -136,8 +136,8 @@
       <% end %>
     </td>
 
-    <td data-label="Average Score" data-score="<%= submission.average_score %>">
-      <%= submission.average_score %>
+    <td data-label="Average Score" data-score="<%= average_score(submission).formatted_score %>">
+      <%= average_score(submission).formatted_score %>
```
stepchud, line 140: I'm seeing an incorrect value for average_score when the submission has a recused evaluation that has a score (I know it's a random edge case, but it doesn't look right in the UI). Probably the right fix is to include only `assigned` evaluations in the Submission model's `Submission#average_score` function
Screenshot 2025-01-07 at 1 03 51 PM

stepchud (changes requested):
one more thing for `average_score` and then I think we can ship

app/views/phases/_submissions_table_rows.html.erb
emmabjj: Okay interesting, yeah that makes sense - I'll go ahead and add that, thanks!

emmabjj: I just noticed the same issue with `in progress` evaluations, so I'll include only `assigned` evaluations with a `completed_at` in the `average_score` function

Screenshot 2025-01-07 at 5 22 03 PM

Screenshot 2025-01-07 at 5 22 31 PM

stepchud (approved):
Looks great! ❤️ 💙 💛 💜

PR #336, opened 2024-12-23
[298] Evaluations routes & controller

### LINKED ISSUE #298: Evaluations routes & controller
# What/Why
In order to view and update eval scores on submission, we need to implement a rails controller to handle the requests.
The scope of this ticket is _backend only_. The view will be completed after.

## Routes
```
resources :evaluations, only: [:new, :edit, :show] do
  member do
    patch :save_draft      
    patch :mark_complete    
  end
end
```

They can save their progress on the evaluation as a draft. They can mark an evaluation complete.

## Acceptance Criteria
- [x] Routes added
- [x] Saving a draft doesn't do validation but it stores the record
- [x] Marking complete validates the scores, and required comments (if they are required on the eval form)

- [ ] WAVE screenshot

## Out of Scope
UI, view templates, rendering the evaluation form, etc

cpreisinger:
#298 

Contains controller actions and specs for creating and editing evaluations and saving them as drafts or making them as complete

Needs future refactor on `evaluations_controller` length possibly moving some functionality into a service

# Other Questions
- Should this prevent creation if their status is recused?

# Out of Scope
- Evaluation form front end
- Total score should get set when evaluation is marked as complete


app/controllers/evaluations_controller.rb
```diff
@@ -3,5 +3,144 @@
...
+
+  def new
+    @evaluator_submission_assignment = find_evaluator_submission_assignment
+
+    if @evaluator_submission_assignment.nil? || !can_access_evaluation?
+      return redirect_to evaluations_path, alert: I18n.t("evaluations.alerts.evaluator_submission_assignment_not_found")
+    end
+
+    phase = @evaluator_submission_assignment.phase
+    @evaluation_form = EvaluationForm.find_by(phase:)
```
stepchud: nit: save a line?
```suggestion
    @evaluation_form = @evaluator_submission_assignment.phase.evaluation_form
```

app/controllers/evaluations_controller.rb
```diff
@@ -3,5 +3,144 @@
...
+    )
+
+    @evaluation_form.evaluation_criteria.each do |criterion|
+      @evaluation.evaluation_scores.build(evaluation_criterion: criterion)
+    end
+  end
+
+  # Redirect Helpers
+  def unauthorized_redirect
+    redirect_to evaluations_path, alert: I18n.t("evaluations.alerts.unauthorized")
```
stepchud, line 131: looks like this translation key is missing
```ruby
(ruby) I18n.t("evaluations.alerts.unauthorized")
"Translation missing: en.evaluations.alerts.unauthorized"
```

stepchud:
Emma added a way to skip individual specs instead of the whole suite. Can you use that instead of skipping all?
```ruby
it 'renders the index page successfully', bullet: :skip do
  # spec throws the bullet unoptimized exception
end
```

config/locales/en.yml
```diff
@@ -87,6 +87,14 @@ en:
   evaluations: 
     unique_user_for_evaluation_form_and_submission_error: "already has an evaluation for this form and submission"
     unique_evaluator_submission_assignment: "already has an evaluation"
+    alerts:
+      evaluator_submission_assignment_not_found: "No assignment found for this submission"
+      evaluation_form_not_found: "No evaluation form found for this submission"
+      save_draft_error:  "Failed to save evaluation as draft: #{errors}."
```
stepchud: typo
```suggestion
      save_draft_error:  "Failed to save evaluation as draft: %{errors}."
```

app/controllers/evaluations_controller.rb
```diff
@@ -3,5 +3,144 @@
...
+
+  def handle_save_draft_failure
+    flash.now[:alert] =
+      I18n.t("evaluations.alerts.save_draft_error", errors: @evaluation.errors.full_messages.to_sentence)
+
+    if @evaluation.new_record?
+      render :new, status: :unprocessable_entity
+    else
+      render :edit, status: :unprocessable_entity
+    end
```
stepchud: we should redirect back to the original url if there are error conditions

config/routes.rb
```diff
@@ -10,7 +10,21 @@
...
+      patch 'mark_complete'
+    end
+    collection do
+      post 'save_draft'
+      post 'mark_complete'
+    end
+  end
+
+  resources :evaluator_submission_assignments, only: [] do
+    resources :evaluations, only: [:new]
```
stepchud: option to nest this route under submissions instead

stepchud (commented):
review feedback mostly from our discussions and a few minor details

app/controllers/evaluations_controller.rb
```diff
@@ -3,5 +3,144 @@
...
+    @evaluator_submission_assignment = find_evaluator_submission_assignment
+
+    unauthorized_redirect unless can_access_evaluation?
+  end
+
+  def find_or_initialize_evaluation
+    if params[:id]
+      Evaluation.includes([:evaluation_criteria]).find(params[:id])
+    else
+      Evaluation.new
```
stepchud: it might be better to attach the current_user to the new record here rather than passing their ID as a hidden param from the form submission
```suggestion
      Evaluation.new(user_id: current_user.id)
```

spec/requests/evaluations_spec.rb
```diff
@@ -173,4 +173,478 @@
...
+        expect(updated_evaluation.additional_comments).to eq("Test")
+
+        expect(updated_evaluation.errors).to be_empty
+        expect(updated_evaluation.completed_at).not_to be_nil
+
+        expect(response).to redirect_to(evaluations_path)
+        expect(flash[:notice]).to include(I18n.t("evaluations.notices.marked_complete"))
+      end
+
+      # TODO: Needs fix
```
stepchud: what needs fix?

app/controllers/evaluations_controller.rb
```diff
@@ -1,9 +1,13 @@
 # frozen_string_literal: true
 
+# TODO: Reenable rubocop after refactor/shortening controller code or moving some functionality into service
+# rubocop:disable all
```
stepchud, line 4: won't block on this, but it's better to disable the specific cop that is failing than all of them. just in case we forget to re-enable or some other changes break different cops than the ClassLength one.

app/controllers/evaluations_controller.rb
```diff
@@ -109,9 +135,9 @@ def handle_save_draft_failure
       I18n.t("evaluations.alerts.save_draft_error", errors: @evaluation.errors.full_messages.to_sentence)
 
     if @evaluation.new_record?
-      render :new, status: :unprocessable_entity
+      redirect_to new_submission_evaluation_path(@evaluator_submission_assignment.submission_id)
```
stepchud, line 139: since the redirect triggers a fresh request/response cycle, you should use `flash[:alert]` instead of flash.now above

app/controllers/evaluations_controller.rb
```diff
@@ -120,9 +146,9 @@ def handle_mark_complete_failure
       I18n.t("evaluations.alerts.mark_complete_error", errors: @evaluation.errors.full_messages.to_sentence)
 
     if @evaluation.new_record?
-      render :new, status: :unprocessable_entity
+      redirect_to new_submission_evaluation_path(@evaluator_submission_assignment.submission_id)
```
stepchud, line 150: update flash.now => flash

spec/requests/evaluations_spec.rb
cpreisinger: I think this was old and I forgot to remove it. When I was having the issue with evaluation_scores failing uniqueness validation

app/controllers/evaluations_controller.rb
```diff
@@ -22,12 +26,148 @@ def submissions
...
+      return redirect_to evaluations_path, alert: I18n.t("evaluations.alerts.evaluation_form_not_found")
+    end
+
+    build_evaluation
+
+    render :new
+  end
+
+  def edit
+    @evaluation = Evaluation.find_by(id: params[:id])
```
stepchud, line 50: this needs to load the evaluator_submission_assignment otherwise it always returns unauthorized (the specs for this request route aren't checking the response status)

stepchud (changes requested):
I'll push a fix for the blockers

stepchud (approved):
looks good to me

PR #340, opened 2024-12-31
[293] Load more Pagination on Submissions list

### LINKED ISSUE #293: Submissions list view data is limited and loads more data on demand
# What/Why
As a challenge manager I want to manage challenges with many public submissions.

This task will limit the initial data load of submissions and add a **Load more** button below the list to view more submissions.

### Figma Link
Screenshot 2024-11-25 at 5 26 11 PM
[Submissions list](https://www.figma.com/design/VfdBFG8ShVnOStdSNYy9fR/Challenge.Gov-Component-Library?node-id=2963-54106&t=KdJ650lG6P1BP9OR-1)

## Technical Notes
- Data should be paginated on the Rails backend and added to the page using Stimulus.
- There is a dependency between #57 / #207 and this ticket. If #207 merges first, pagination will need to work for sort and filter options as well as the default view.

## Acceptance Criteria
- [x] Page initially loads up to 20 submissions
- [x] **Load more** button loads 20 more records
- [x] Pagination happens in SQL queries, not in memory

- [x] WAVE screenshot

emmabjj:
Related ticket: #293 

Limit the initial data load of submissions and add a **Load more** button below the list to view more submissions. Data is _paginated_ on the Rails backend

This feature depended on the completion of #207 & #57 (Sort & Filter). 

Screenshot 2024-12-30 at 6 22 14 PM

#### Tablet & Mobile
Screenshot 2024-12-30 at 6 29 32 PM
Screenshot 2024-12-30 at 6 22 33 PM

#### WAVE
Screenshot 2024-12-30 at 6 23 06 PM
Screenshot 2024-12-30 at 6 23 19 PM

#### Interaction with Sort & Filter
load_more_submissions

### DIFF
--- a/app/javascript/controllers/load_more_controller.js
+++ b/app/javascript/controllers/load_more_controller.js
@@ -1,0 +1,40 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static targets = ["container", "loadMoreButton"]
+  static values = { 
+    page: Number,
+    totalCount: Number
+  }
+
+  connect() {
+    this.pageValue = 1
+    this.checkHasMoreSubmissions()
+  }
+
+  async loadMore() {
+    this.pageValue++
+    
+    try {
+      const url = new URL(window.location.href)
+      url.searchParams.set('page', this.pageValue)
+      url.searchParams.set('format', 'json')
+      
+      const response = await fetch(url)
+      if (!response.ok) throw new Error('Network response was not ok')
+      
+      const html = await response.text()
+      this.containerTarget.insertAdjacentHTML('beforeend', html)
+      this.checkHasMoreSubmissions()
+    } catch (error) {
+      console.error("Error loading more submissions:", error)
+    }
+  }
+
+  checkHasMoreSubmissions() {
+    const rows = this.containerTarget.querySelectorAll('tr')
+    if (rows.length >= this.totalCountValue) {
+      this.loadMoreButtonTarget.classList.add('display-none')
+    }
+  }
+}
--- a/app/javascript/controllers/load_more_controller.js
+++ b/app/javascript/controllers/load_more_controller.js
@@ -18,9 +18,14 @@ export default class extends Controller {
     try {
       const url = new URL(window.location.href)
       url.searchParams.set('page', this.pageValue)
-      url.searchParams.set('format', 'json')
+      url.searchParams.set('partial', 'true')
       
-      const response = await fetch(url)
+      const response = await fetch(url, {
+        headers: {
+          'Accept': 'text/html',
+          'X-Requested-With': 'XMLHttpRequest'
+        }
+      })
       if (!response.ok) throw new Error('Network response was not ok')
       
       const html = await response.text()
--- a/app/views/phases/_submissions_table.html.erb
+++ b/app/views/phases/_submissions_table.html.erb
@@ -8,7 +8,11 @@
 
 <%= render partial: "phases/sort_and_filter" %>
 
-<table class="usa-table usa-table--stacked-header usa-table--borderless width-full gray-header">
+<div data-controller="load-more" 
+     data-load-more-url-value="<%= submissions_phase_path(@phase, format: :json) %>"
+     data-load-more-total-count-value="<%= @filtered_count %>">
+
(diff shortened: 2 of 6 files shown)


stepchud:
please note there was a conflict with #313 which I resolved by rebasing this branch so you will need to pull in this branch locally if you want to make any further changes, otherwise you will have the same conflicts.

app/views/phases/_submissions_table.html.erb
```diff
@@ -8,7 +8,11 @@
 
 <%= render partial: "phases/sort_and_filter" %>
 
-<table class="usa-table usa-table--stacked-header usa-table--borderless width-full gray-header">
+<div data-controller="load-more" 
+     data-load-more-url-value="<%= submissions_phase_path(@phase, format: :json) %>"
```
stepchud: is this value being used?

app/javascript/controllers/load_more_controller.js
```diff
@@ -0,0 +1,40 @@
...
+      const url = new URL(window.location.href)
+      url.searchParams.set('page', this.pageValue)
+      url.searchParams.set('format', 'json')
+      
+      const response = await fetch(url)
+      if (!response.ok) throw new Error('Network response was not ok')
+      
+      const html = await response.text()
+      this.containerTarget.insertAdjacentHTML('beforeend', html)
+      this.checkHasMoreSubmissions()
```
stepchud: this is a creative approach and it seems to work correctly. one thing I noticed is that you are requesting json formatted data, but the server returns an html partial. it's odd, but mostly a web standards violation, and maybe not that critical. one way to workaround might be to request regular html response. instead of responding to json in the controller, check for the `page` param or maybe some other custom param like `partial=true` and then use that to conditionally render the full page or just the partial result of the pagination query.
the controller might then look something like this:
```ruby
    respond_to do |format|
      format.html do
        if params[:page]
          render partial: 'submissions_table_rows',
            locals: { submissions: @submissions },
            formats: [:html]
        else
          render :submissions
      end
```
another approach might be to create a json partial and render json data to the stimulus controller which would then map the json data to html elements and insert those into the page using javascript. I'm not sure how that would impact performance compared to inserting the whole fragment at once like you are now. one benefit of rendering json is that you could add other data to the json response, such as an optional link to the next page, instead of counting the number of elements on the page and comparing with the total count to dynamically display the "load more" button. you could just not send a link to the next page in the response if there is no more data, if that makes sense. there are a few json libraries that take this approach.

if you do decide to convert to json response, it might be good to go all way and refactor the page so that the initial page loads an empty list and then fetch *all* submissions data, including the initial rows, via json requests from stimulus.

let me know what you think about that or if you want to discuss the options in person.

spec/requests/submissions_spec.rb
```diff
@@ -233,5 +233,65 @@
...
+    context 'when paginating submissions' do
+      let!(:submissions) do
+        (1..25).map do |n|
+          create(:submission, challenge: challenge, phase: phase)
+        end
+      end
+
+      it 'returns first page of submissions' do
+        get submissions_phase_path(phase)
+        expect(response.body).to have_css('tr[data-submission-id]', count: 20)
```
stepchud, line 246: want to add an expectation that the "Load more" button exists and is hidden when there is no more data?

stepchud (approved):
I'm ok with this as is, I think it's working properly. I left some non-blocking feedback if you want to check that out.

app/javascript/controllers/load_more_controller.js
stepchud: to be clear, I think it's okay as is. out of these options I would probably go with a custom param returning partial html because it seems like the quicker fix and doesn't rely on an existing param (`page`) which might be more flexible and less bug prone.

app/views/phases/_submissions_table.html.erb
emmabjj: It was being used when I had a different approach, it's no longer being used - I'll remove it, thanks!

app/javascript/controllers/load_more_controller.js
emmabjj: Okay yes, this makes sense, thanks for explaining this! I also prefer going with a custom param returning partial html, I'll go ahead and make that change. I had a couple different approaches I was deciding between, but this approach is cleanest!

PR #353, opened 2025-01-10
remove unnecessary association that is breaking things

stepchud:
There is an issue with #292 where submissions with `manager_id: nil` can't be updated. However, this is not necessary and the association can be disregarded.
The issue discovered on staging prevented submissions from being marked eligible for evaluation because they were missing the "required" `belongs_to :manager`, which was not actually required.

### DIFF
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -33,7 +33,6 @@ class Submission < ApplicationRecord
   belongs_to :challenge
   belongs_to :phase, counter_cache: true
   belongs_to :submitter, class_name: 'User'
-  belongs_to :manager, class_name: 'User'
   has_many :evaluator_submission_assignments, dependent: :destroy
   has_many :evaluators, through: :evaluator_submission_assignments, class_name: "User"
   has_many :evaluations, dependent: :destroy
--- a/app/views/phases/_submissions_stats.html.erb
+++ b/app/views/phases/_submissions_stats.html.erb
@@ -11,9 +11,9 @@
                 <span class="font-sans-3xl text-primary text-bold"><%= @submissions_count %></span>
             </div>
             <div>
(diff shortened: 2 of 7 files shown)


app/models/submission.rb
```diff
@@ -33,7 +33,6 @@ class Submission < ApplicationRecord
   belongs_to :challenge
   belongs_to :phase, counter_cache: true
   belongs_to :submitter, class_name: 'User'
-  belongs_to :manager, class_name: 'User'
```
stepchud, line 36: Rails `belongs_to` associations are required by default. We could add the `optional: true` config option to the association to make it optional but I decided to remove it because I can't find any reason we need to have the manager associated with the submission and this field should probably be removed eventually. I'm not going to do that migration right now.

This is causing an issue on staging because in the real world submissions are created without the manager_id which causes those records to fail validations on other kinds of updates in Rails.

stepchud: This is causing an issue on staging because in the real world submissions are created without the `manager_id` which causes those records to fail validations on other kinds of updates in Rails.

app/views/phases/_submissions_stats.html.erb
```diff
@@ -11,9 +11,9 @@
                 <span class="font-sans-3xl text-primary text-bold"><%= @submissions_count %></span>
             </div>
             <div>
-                <h3 class="usa-summary-box__heading text-primary margin-bottom-05" id="summary-box-key-information">
+                <h2 class="usa-summary-box__heading text-primary margin-bottom-05" id="summary-box-key-information">
```
stepchud, line 14: these changes fix the a11y violations on the Phase Submissions page

cpreisinger (approved):

PR #358, opened 2025-01-14
[1508] modify migration to succeed if the column already exists

stepchud:
Related to #1508, this updates the existing Rails migration to make it optional if the migration already ran on the Phoenix app.

Based on the Rails API documentation - [add_column](https://api.rubyonrails.org/v8.0.1/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-add_column) options


emmabjj (approved):
Looks good to me 👍

Issue #285, opened 2024-11-20
Submissions on the Rails App - Evaluators


r-bartlett-gsa:
### User story
As a challenge manager or evaluator, in order to manager evaluations process efficiently, I would like to be able to access submissions on the same app as Evaluation Forms.

**Acceptance Criteria:**

- [x] A user can navigate to a page on the Rails app that lists challenges via the dashboard or utility nav:
  - [x] An evaluator sees a list of challenges they have assigned submission to, and the list displays:
  - [x] Challenge Name and Phase
  - [x] Number of assigned submissions
  - [x] Number of remaining submissions to evaluate (in progress and not started)
  - [x] Evaluations due date
  - [x] If the user does not have any submissions assigned to them, a message is displayed indicating so
- [x] The user can view a list of submission for each challenge:
  - [x] https://github.com/GSA/Challenge_platform/issues/82

Dev tasks:

- [ ] Task 1
- [ ] Task 2

### Definition of Done

**Doing (dev team)**
- [x] Code complete 
- [x] Code is organized appropriately
- [x] Any known trade offs are documented in the associated GH issue
- [x] Code is documented, modules, shared functions, etc.
- [x] Automated testing has been added or updated in response to changes in this PR
- [x] The feature is smoke tested to confirm it meets requirements
- [x] Database changes have been peer reviewed for index changes and performance bottlenecks
- [x] PR that changes or adds UI
  - [x] include a screenshot of the WAVE report for the altered pages
  - [x] Confirm changes were validated for mobile responsiveness
- [x] PR approved / Peer reviewed 
- [x] Security scans passed
- [x] Automate accessibility tests passed
- [x] Build process and deployment is automated and repeatable
- [x] Feature toggles if appropriate
- [x] Deploy to staging
- [x] Move card to testing column in the board
---
**Staging**
- [x] Usability testing: mobile and desktop (Tracy or Marni)
- [ ] ~~Accessibility tested (Marni)~~
  - [ ]  ~~Keyboard navigation~~
  - [ ]  ~~Focus confirmed~~
  - [ ] ~~Color contrast compliance~~
  - [ ] ~~Screen reader testing~~
- [x] Cross browser testing - UI rendering is performant on below listed devices/browsers (Tracy or Marni)
  - [x] Windows/Chrome
  - [x] Windows/Edge
  - [x] Mac/Chrome
  - [x] Mac/Safari
  - [x] iOS/Safari
- [x] AC review (Renata)
[...]

PR #359, opened 2025-01-15
[285] Adjust Challenge Phase index query for Evaluators

emmabjj:
Related ticket: #285 

The list of challenges should only show the challenges that the evaluator has assigned submissions to.

Screenshot 2025-01-14 at 7 24 00 PM
Screenshot 2025-01-14 at 7 24 23 PM

### DIFF
--- a/app/views/evaluations/_phases_table.html.erb
+++ b/app/views/evaluations/_phases_table.html.erb
@@ -33,8 +33,6 @@
               View Submissions
             <% end %>
           </div>
-          </div>  
-        </td>
         </td>  
       </tr>
     <% end %>
--- a/app/views/evaluations/_phases_table.html.erb
+++ b/app/views/evaluations/_phases_table.html.erb
@@ -9,36 +9,34 @@
     </tr>
   </thead>
   <tbody>
-    <% @challenges.each do |challenge| %>
-      <% challenge.phases.each do |phase| %>
-        <tr>
-          <th data-label="Challenge Title" scope="row" class="text-top">
-            <%= challenge_phase_title(challenge, phase) %>
-          </th>
-          <td data-label="Assigned to me" class="text-top">
-            <%= assigned_submissions_count(current_user, challenge, phase) %>
-          </td>
-          <td data-label="Remaining to evaluate" class="text-top">
-            <%= remaining_evaluations_count(current_user, challenge, phase) %>
-          </td>
-          <td data-label="Due by" class="text-top">
-            <% if phase.evaluation_form %>
-              <%= phase.evaluation_form.closing_date %>
-            <% else %>
-              N/A
+    <% @phases.each do |phase| %>
+      <tr>
+        <th data-label="Challenge Title" scope="row" class="text-top">
+          <%= challenge_phase_title(phase.challenge, phase) %>
+        </th>
+        <td data-label="Assigned to me" class="text-top">
+          <%= assigned_submissions_count(current_user, phase.challenge, phase) %>
+        </td>
+        <td data-label="Remaining to evaluate" class="text-top">
+          <%= remaining_evaluations_count(current_user, phase.challenge, phase) %>
+        </td>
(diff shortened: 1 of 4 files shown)


spec/requests/evaluations_spec.rb
```diff
@@ -62,6 +62,43 @@
...
+
+      before do
+        ChallengePhasesEvaluator.create!(
+          challenge: challenge,
+          phase: phase,
+          user: evaluator
+        )
+      end
+
+      it "only shows challenges with assigned submissions" do
```
stepchud: 👍 can we add some specs for including `status: :recused` submissions and excluding `status: :unassigned` and `status: :recused_unassiged`?

emmabjj: Sure!

emmabjj: @stepchud I added `is_multi_phase: false` to the spec to prevent the challenge factory from creating 3 phases for the challenge created. This is so the spec doesn't return challenge phases with no assigned submissions, which ends up looking like the behavior we're testing to avoid. 👍

spec/requests/evaluations_spec.rb
```diff
@@ -62,6 +62,61 @@
...
+
+        get evaluations_path
+
+        expect(response.body).to include(challenge.title)
+        expect(response.body).not_to include(other_challenge.title)
+        expect(response.body.scan(/data-label="Challenge Title"/).count).to eq(1)
+        expect(response.body).to have_css('td[data-label="Assigned to me"]')
+        expect(response.body).to have_css('td[data-label="Assigned to me"]', text: '2')
+        expect(response.body).to have_css('td[data-label="Remaining to evaluate"]')
+        expect(response.body).to have_css('td[data-label="Remaining to evaluate"]', text: '2')
```
stepchud: nit: these are duplicated expectations and can be collapsed to one line
```suggestion
        expect(response.body).to have_css('td[data-label="Assigned to me"]', text: '2')
        expect(response.body).to have_css('td[data-label="Remaining to evaluate"]', text: '2')
```

spec/requests/evaluations_spec.rb
stepchud: if the challenge is_muli_phase, and the evaluator is only assigned to one of the phases, we only want it to show that assigned phase in the list. probably the view should query for assigned/recused phases instead of challenges so that you can properly filter them? otherwise this isn't the correct behavior.

stepchud (changes requested):
needs to filter for assigned phases on is_multi_phase challenges

spec/requests/evaluations_spec.rb
stepchud: for example, here I have a multi-phase challenge and only assigned this evaluator to one of the phases, so they should only see the first and second row in this list, not the last two:

Screenshot 2025-01-15 at 8 59 33 AM

emmabjj: Yep, that makes sense - I'll fix the query and update the tests, thanks!

emmabjj:
@stepchud I've adjusted the query so it queries on phases in which that evaluator has assigned or recused submissions instead of challenges. 

The updated test now includes a multi-phase challenge. It checks that the evaluator only sees the phase of the multi-phase challenge in which they have assigned/recused submissions. And it still includes the expectations around including `status: :recused` submissions and excluding `status: :unassigned` and `status: :recused_unassigned` submissions. 

Lemme know what you think! Thanks!

app/views/evaluations/_phases_table.html.erb
```diff
@@ -9,36 +9,34 @@
...
-            </div>
-            </div>  
-          </td>
-          </td>  
-        </tr>
-      <% end %>  
-    <% end %>  
+          </div>
+          </div>  
+        </td>
```
stepchud: remove duplicate closing tags
```suggestion
```

stepchud (approved):

PR #369, opened 2025-01-23
[284] Submission Document Access

### LINKED ISSUE #284: Uploaded Files access
# What/Why
As a Challenge Manage/Evaluator, I would like to view the uploaded files for a submission I am managing/evaluating.
The Submission Details view should include a list of links that the user can download files from.

## Technical Notes
We should investigate the access permissions on the phoenix app in case there are additional requirements.

### Figma Link
[Submission Details](https://www.figma.com/design/VfdBFG8ShVnOStdSNYy9fR/Challenge.Gov-Component-Library?node-id=2986-66670&t=LXClgrijJuwuc6A0-4)

## Acceptance Criteria
- [ ] If a submission has uploads, they appear on the detail view
- [ ] Clicking the link downloads the correct file

- [ ] WAVE screenshot

stonefilipczak:
Allows access to a submission's uploaded documents from the submission details page. 

Adds the concept of documents to the Rails application and adds the UI for viewing them.

Clicking a document's link will open the document in a new tab where it can be viewed or downloaded. 

Urls are to local /uploads folder for dev & test environments and to an s3 bucket for staging and prod. 

 
Screen Shot 2025-01-23 at 3 21 40 PM


stepchud (changes requested):
Although I haven't been able to fully test it out yet, it seems like this solution could be a workaround for local development. However it will not work when deployed since staging and production ENVs store files on S3 and this PR hardcodes the path to an uploads directory on the Phoenix app.
In the past I've used the Paperclip gem for this type of thing (attaching files to Rails models) but now that is deprecated in favor of [ActiveStorage](https://guides.rubyonrails.org/active_storage_overview.html). I think this should be the preferred approach and we can configure Rails to use the same S3 bucket as Phoenix.

stonefilipczak:
> Although I haven't been able to fully test it out yet, it seems like this solution could be a workaround for local development. However it will not work when deployed since staging and production ENVs store files on S3 and this PR hardcodes the path to an uploads directory on the Phoenix app. In the past I've used the Paperclip gem for this type of thing (attaching files to Rails models) but now that is deprecated in favor of [ActiveStorage](https://guides.rubyonrails.org/active_storage_overview.html). I think this should be the preferred approach and we can configure Rails to use the same S3 bucket as Phoenix.

Ah gotcha, didn't realize those production architectural concerns, I figured replicating what happens on the phoenix app would work for this. Will pick it back up Monday!

stepchud:
> Although I haven't been able to fully test it out yet, it seems like this solution could be a workaround for local development. However it will not work when deployed since staging and production ENVs store files on S3 and this PR hardcodes the path to an uploads directory on the Phoenix app. In the past I've used the Paperclip gem for this type of thing (attaching files to Rails models) but now that is deprecated in favor of [ActiveStorage](https://guides.rubyonrails.org/active_storage_overview.html). I think this should be the preferred approach and we can configure Rails to use the same S3 bucket as Phoenix.

Just to update this discussion thread, we decided not to implement the full ActiveStorage in favor of a more focused approach that uses the [Fog gem](https://github.com/fog/fog-aws?tab=readme-ov-file#s3) to render signed AWS links for direct download instead. ActiveStorage would require a much larger migration that impacts Phoenix and not worth it at this time to satisfy the access to Submission attachments.

Gemfile.lock
```diff
@@ -144,14 +151,150 @@ GEM
...
+    fog (1.38.0)
+      fog-aliyun (>= 0.1.0)
+      fog-atmos
+      fog-aws (>= 0.6.0)
+      fog-brightbox (~> 0.4)
+      fog-cloudatcost (~> 0.1.0)
+      fog-core (~> 1.32)
+      fog-dynect (~> 0.0.2)
+      fog-ecloud (~> 0.1)
+      fog-google (<= 0.1.0)
```
stepchud: this brings in a lot of dependencies for the ability to generate an s3 signed url, can you try using the [fog-aws gem](https://github.com/fog/fog-aws) instead to reduce the bloat here? (in the Gemfile). btw I assume this is a straightforward swap and replace, if this takes a lot more work please let me know.
```
- gem "fog", "~> 1.38"
+ gem "fog-aws"
```

app/models/document.rb
```diff
@@ -21,9 +21,29 @@ class Document < ApplicationRecord
   belongs_to :submission
 
   def external_url
+    if ENV["RAILS_ENV"] == "production" || ENV["RAILS_ENV"] == "staging"
```
stepchud: I think the standard way to do this is `Rails.env.production? || Rails.env.staging?`

app/models/document.rb
```diff
@@ -21,9 +21,29 @@ class Document < ApplicationRecord
...
+      prod_external_url
+    else
+      dev_external_url
```
stepchud: instead of naming these after an environment, can we call them `s3_link` and `file_system_link`?

stepchud (approved):
LGTM. I tested the S3 attachment by copying one from the eval-dev cloud environment to my local machine and forcing the app use use s3 instead of file storage. I also tested file system storage is still working locally when not forced to use s3. 👍

PR #380, opened 2025-01-30
[370] Hide Soft Deleted Submissions

### LINKED ISSUE #370: Deleted submissions
When a submission is deleted it currently is only reflected in the submission count on the challenges list table, but not on the submissions list table or progress dashboard (and also not in the phoenix app) A challenge manager can still review a deleted submission.

See screencast https://app.screencast.com/mw8oIiqZNLZle where submission ID 13 was deleted.
Phoenix app: https://app.screencast.com/AmDCgVHnvkvCE

Acceptance criteria:

- [x] When a submission is deleted 
  - [x] the submission count is updated in all sections of the site
  - [x] the submission materials cannot be accessed by the challenge manager

stonefilipczak:
Filters soft-deleted submissions out of the all_submissions association recently added to phases. This hides deleted submissions from most places they are represented in the application. 

The remaining situations rely on a rails-provided counter cache which shows a submission count on the phase. This needed to be modified to filter out deleted submissions as well. I added the counter_culture gem to allow for this behavior and set up an implementation which appears to work.

My only concern is wether or not the counter cache values would have to be manually updated every time a submission is deleted on the phoenix side. My local experiments have been confounding. Counter culture gem provides a method for manually updating counts which we would use like `Phase.counter_culture_fix_counts`, see the [documentation](https://www.rubydoc.info/gems/counter_culture/0.1.33#manually-populating-counter-cache-values)

I request that others test locally and see how it behaves for them.


app/models/phase.rb
```diff
@@ -23,7 +23,9 @@ class Phase < ApplicationRecord
   belongs_to :challenge
   # More relations from phoenix app
   has_many :all_submissions, class_name: "Submission", dependent: :destroy
-  has_many :submissions, -> { where(status: "submitted") }, inverse_of: :phase, dependent: :destroy
+  has_many :submissions, lambda {
+    where(status: "submitted").where(deleted_at: nil)
+  }, inverse_of: :phase, dependent: :destroy
```
stepchud, line 28: I think this is enough for the current counter_cache column to work?

app/models/submission.rb
```diff
@@ -64,6 +64,7 @@ class Submission < ApplicationRecord
     else
       none
     end
+    by_user_role.where(deleted_at: nil)
```
stepchud, line 68: added filtering here for this cope that is used to give users access to submissions based on their role

stepchud (approved):
After reviewing this, I don't think the counter_culture gem is necessary and I'm not sure we need two columns tracking submission counts. The phoenix app seems to work okay as far as updating the current counter when a submission is deleted, so for this bug we just need to make sure that filtering is done on the rails side too. I reverted some of your changes, and added some specs. 👍

Issue #75, opened 2024-07-31
Assign Evaluators to a Submission - Manual


r-bartlett-gsa:
### User story

As a challenge manager, in order to have the submission of my challenge reviewed and scored by evaluators, I would like to be able to assign evaluators to submissions.

**Acceptance criteria:**

- [x] Under each submission detail view, a challenge manager can assign as many evaluators as needed from the the available evaluators list
- [x] Evaluators can only be assigned if the submissions is checked as eligible for evaluations / is prescreened
- [x] Only users that meet the following criteria can be assigned to the submission of the challenge:
  - [x] User with evaluator role and approved accounts
  - [x] Evaluators that are part of the evaluator list for that challenge
  - [x] Evaluators that are not recused
  - [ ] ~Evaluators that are not unassigned~
- [x] A challenge manager has an option to review their challenge evaluator list by clicking on the link button at the bottom of the available evaluators list
- [x] Evaluators evaluation progress is tracked for that submission
- [x] When an evaluator is assigned to a submission, that submission appears on the evaluator's list of submissions #82  
- [x] When an evaluator is assigned to a submission, that evaluator appears on the challenge manager's list of submissions #59 

**Dev tasks:**

- [x] Add `create` route for evaluator_submission_assignments
- [x] #344 

### Definition of Done

**Doing (dev team)**
- [x] Code complete 
- [x] Code is organized appropriately
- [x] Any known trade offs are documented in the associated GH issue
- [x] Code is documented, modules, shared functions, etc.
- [x] Automated testing has been added or updated in response to changes in this PR
- [x] The feature is smoke tested to confirm it meets requirements
- [x] Database changes have been peer reviewed for index changes and performance bottlenecks
- [x] PR that changes or adds UI
  - [x] include a screenshot of the WAVE report for the altered pages
  - [x] Confirm changes were validated for mobile responsiveness
- [x] PR approved / Peer reviewed 
- [x] Security scans passed
- [x] Automate accessibility tests passed
- [x] Build process and deployment is automated and repeatable
- [x] Feature toggles if appropriate
- [x] Deploy to staging
- [x] Move card to testing column in the board
---
**Staging**
- [x] Accessibility tested (Marni)
  - [x]  Keyboard navigation
  - [x]  Focus confirmed
  - [x] Color contrast compliance
  - [x] Screen reader testing 
- [x] Usability testing: mobile and desktop (Tracy or Marni)
[...]

PR #384, opened 2025-02-04
[344] Enforce uniqueness of evaluator submission assignments.

### LINKED ISSUE #344: Enforce uniqueness of evaluator_submission_assignment
# What/Why
As a Challenge Manager I want to ensure that submissions are only assigned to evaluators once. It should not be possible to create a duplicate assignment record with the same user_id and submission_id.

## Acceptance Criteria
- [x] `evaluator_submission_assignments` have a unique index on user_id/submission_id
  - [x] Model validation adds an error message if a duplicate submission assignment occurs

stonefilipczak:
Adds a model validation to Evaluator Submission Assignments which validates uniqueness of submission ID against user ID.

Also adds a unique index to the database for another layer of protection against duplicates. 

As a bonus, makes a phase's list of evaluators only return users with the role "evaluator", thus excluding other user roles from evaluator submission assignment.


db/structure.sql
```diff
@@ -2465,7 +2473,9 @@ ALTER TABLE ONLY public.winners
 SET search_path TO "$user", public;
 
 INSERT INTO "schema_migrations" (version) VALUES
+(20250204131151),
 (20250202220815),
+(20250130144731),
```
stepchud: looks like this line could be breaking the migration since it was removed from your other PR
```suggestion
```

db/structure.sql
```diff
@@ -843,7 +843,8 @@ CREATE TABLE public.phases (
     how_to_enter_delta text,
     inserted_at timestamp(0) without time zone NOT NULL,
     updated_at timestamp(0) without time zone NOT NULL,
-    submissions_count integer DEFAULT 0 NOT NULL
+    submissions_count integer DEFAULT 0 NOT NULL,
+    active_submissions_count integer DEFAULT 0 NOT NULL
```
stepchud: this column needs to be removed since it is no longer part of the schema. you can drop it directly from your database, or create and run a migration that drops it and then delete the migration since it won't be committed/merged.
```suggestion
```

app/models/phase.rb
```diff
@@ -29,7 +29,7 @@ class Phase < ApplicationRecord
   # has_one :winner, class_name: 'PhaseWinner'
   has_many :evaluator_invitations, dependent: :destroy
   has_many :challenge_phases_evaluators, dependent: :destroy
-  has_many :evaluators, through: :challenge_phases_evaluators, source: :user
+  has_many :evaluators, -> { where(role: "evaluator") }, through: :challenge_phases_evaluators, source: :user
```
stepchud: there is an issue here if I add a non-evaluator role user to the phase, they will not be shown in the list of evaluators. for example, I invited a user who is currently a solver to be an evaluator but they don't show up in the list below:
Screenshot 2025-02-04 at 10 18 54 AM
for #75 we just want to filter the list of available evaluators so that users can't be assigned to submissions until their role is updated to `evaluator`. (this list)
Screenshot 2025-02-03 at 7 49 14 AM

stepchud: take a look at [Submission#available_evaluators](https://github.com/GSA/Challenge_platform/blob/dev/app/models/submission.rb#L89-L92) to filter that list

stepchud (changes requested):
let's filter the available_evaluators list by the evaluator role. can you also add a few request specs that show: non-evaluator role users appear in the "Manage Evaluators" list and only evaluator-role users appear in the "Available Evaluators" list? (the spec should currently fail without the filtering)

db/structure.sql
```diff
@@ -2465,6 +2472,8 @@ ALTER TABLE ONLY public.winners
 SET search_path TO "$user", public;
 
 INSERT INTO "schema_migrations" (version) VALUES
+(20250205135218),
```
stepchud: extraneous version
```suggestion
```

stepchud (approved):
lgtm

Issue #112, opened 2024-08-07
Delete Evaluators


r-bartlett-gsa:
### User story

As a challenge manager, in order to maintain an accurate list of available evaluators, I would like to be able to delete an evaluator from my list.

**Acceptance criteria:**

- [x] A challenge manager can delete evaluators that they entered for their challenge
  - [x] Deleted evaluators are removed from the evaluator list for that challenge and they are not available for submissions assignment of that challenge
  - [x] If an evaluator that is being deleted is already assigned to a submission, a message is displayed for a challenge manager indicating so and asking to confirm deletion
    - [x] Deleted evaluator is removed from assignment to submissions
    - [x] Evaluation data saved or submitted by a deleted evaluator is not displayed and is not used for average score calculation

**Dev tasks:**

- [x] #342

### Definition of Done

**Doing (dev team)**
- [x] Code complete 
- [x] Code is organized appropriately
- [x] Any known trade offs are documented in the associated GH issue
- [x] Code is documented, modules, shared functions, etc.
- [x] Automated testing has been added or updated in response to changes in this PR
- [x] The feature is smoke tested to confirm it meets requirements
- [x] Database changes have been peer reviewed for index changes and performance bottlenecks
- [x] PR that changes or adds UI
  - [x] include a screenshot of the WAVE report for the altered pages
  - [x] Confirm changes were validated for mobile responsiveness
- [x] PR approved / Peer reviewed
- [x] Security scans passed
- [x] Automate accessibility tests passed
- [x] Build process and deployment is automated and repeatable
- [x] Feature toggles if appropriate
- [ ] ~Deploy to staging~ 
- [x] Move card to testing column in the board
---
**Staging**
- [x] Accessibility tested (Marni)
  - [x]  Keyboard navigation
  - [x]  Focus confirmed
  - [x] Color contrast compliance
  - [x] Screen reader testing 
- [x] Usability testing: mobile and desktop (Tracy or Marni)
- [x] Cross browser testing - UI rendering is performant on below listed devices/browsers (Tracy or Marni)
  - [x] Windows/Chrome
  - [x] Windows/Edge
  - [x] Mac/Chrome
  - [x] Mac/Safari
  - [x] iOS/Safari
- [x] AC review (Renata)
- [ ] ~Deploy to production (production-like environment for eval capability) (dev team)~
- [x] Move to production column in the board
---
**Production**
- [ ] User and security documentation has been reviewed for necessary updates (Renata / Tracy / Dev team) 
[...]

PR #404, opened 2025-02-07
[342] Delete assignments and evaluations when an evaluator is deleted

### LINKED ISSUE #342: Delete assignments and evaluations when an evaluator is deleted
# What/Why
When an Evaluator is removed from a Challenge phase, we need to remove the other associations if they exist.
Currently the `EvaluatorManagementService` is only deleting the `ChallengePhaseEvaluator`, however it should also delete any `EvaluatorSubmissionAssignment` and `Evaluation` records associated with that user, if any exist.
NOTE: there is already a modal that warns the user this will happen, but the backend is not implemented.

## Acceptance Criteria
1. Create an evaluator user on a challenge phase
2. Assign them to a submission
3. Complete an evaluation from the user
4.  Delete the evaluator from the challenge phase
- [x] The EvaluatorSubmissionAssignment should be deleted from the DB.
- [x] The Evaluation should be deleted from the DB.

- [ ] WAVE screenshot

emmabjj:
Linked ticket: #112 #342 

AC: When an Evaluator is removed from a Challenge phase, we need to remove the other associations if they exist.
Currently the EvaluatorManagementService is only deleting the ChallengePhaseEvaluator, however it should also delete any EvaluatorSubmissionAssignment and Evaluation records associated with that user, if any exist.

Note: Given that EvaluatorSubmissionAssignment has `has_one :evaluation, dependent: :destroy`, when destroy_all is called on the assignments, it will trigger the dependent destroy callback and delete the associated evaluation as well.

### DIFF
--- a/app/services/evaluator_management_service.rb
+++ b/app/services/evaluator_management_service.rb
@@ -101,15 +101,17 @@ class EvaluatorManagementService
   end
 
   def remove_user_evaluator(evaluator_id)
-    evaluator = User.find(evaluator_id)
-    cpe = ChallengePhasesEvaluator.find_by(challenge: @challenge, phase: @phase, user: evaluator)
-    if cpe.destroy
-      { success: true, message: I18n.t('evaluators.remove_user_evaluator.success') }
-    else
-      { success: false, message: I18n.t('evaluators.remove_user_evaluator.failure') }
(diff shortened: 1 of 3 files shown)


stepchud (approved):
looks good!

PR #429, opened 2025-02-19
[418] Sort submission by number of assignees

### LINKED ISSUE #418: Sort by: Number of assigned evaluators
# What/Why
As a challenge manager, I want to sort the submissions of a phase by the number of assigned evaluators (ascending and descending)

### Figma
Image

## Acceptance Criteria
- [x] `Assigned evaluators: high to low` sorts the submissions by those with the most assigned evaluators first
- [x] `Assigned evaluators: low to high` sorts the submissions by those with the least assigned evaluators first

- [x] WAVE screenshot

stepchud:
Fixes #418 

WIP Status:
- [x] create the `order_by_assignee_count` query
- [x] add model specs for the query
- [x] allow `PhasesController` to sort and filter all options efficiently
  - see code/PR comment
- [x] write request specs for the new sort options (placeholder specs have been written)


app/controllers/phases_controller.rb
```diff
@@ -10,7 +10,9 @@ def index
   end
 
   def submissions
-    @submissions = @phase.submissions.includes(evaluator_submission_assignments: [:evaluator, :evaluation])
+    # This causes a SQL error when sorting by assignee_count -- refactor into the service??
+    # @submissions = @phase.submissions.includes(evaluator_submission_assignments: [:evaluator, :evaluation])
+    @submissions = @phase.submissions
```
stepchud: we should try to preserve the `includes` for the other filtered views, or resolve the SQL error for the assignee sort view. not sure the best way, but maybe we can apply the `includes` inside the SubmissionsSortAndFilterService since neither `set_submission_counts` or `set_submission_statuses` seem to rely on the included models. let's try that.

app/services/submissions_sort_and_filter_service.rb
```diff
@@ -77,4 +78,11 @@ def apply_sorting
       @submissions = @submissions.order(id: :asc)
     end
   end
+
+  def apply_includes
+    @submissions = @submissions.public_send(
+      @params[:sort]&.include?('assignees') ? :preload : :includes,
+      evaluator_submission_assignments: [:evaluator, :evaluation]
+    )
```
stepchud: I would avoid `public_send` if it's not needed, it's better to be explicit (see suggestion).  Maybe the conditional isn't necessary and we can use `preload` for both cases though? 🤔 
```suggestion
    @submissions = if @params[:sort]&.include?('assignees')
      @submissions.preload(evaluator_submission_assignments: [:evaluator, :evaluation])
    else
      @submissions.includes(evaluator_submission_assignments: [:evaluator, :evaluation])
    end
```

stepchud: I saw a Bullet unoptimized query locally and tested this, it seems to fix the optimization for all sort/filter params
```suggestion
    @submissions = @submissions.preload(
      evaluator_submission_assignments: [:evaluator, {evaluation: :evaluation_scores} ]
    )
```

stepchud (commented):
@emmabjj sorry for the late request but I was asking Renata about this PR. She would like us to filter out submissions that are not `eligible_for_evaluation` from both "Assigned evaluators:" sorts, so that the sort only shows submissions that have been selected for evaluation.
In addition to that she would like to filter both "Average score:" to show only those submissions where evaluation is completed (since incomplete submissions shouldn't have scores anyways).
I think both of these should be fairly simple to execute since it just adds an extra scope to each of them plus fixing specs.

app/services/submissions_sort_and_filter_service.rb
```diff
@@ -67,10 +68,24 @@ def apply_sorting
...
   end
+
+  def apply_includes
+    association_preloads = if @params[:sort]&.include?('assignees')
+                             { evaluator_submission_assignments: [:evaluator, { evaluation: :evaluation_scores }] }
+                           else
+                             { evaluator_submission_assignments: [:evaluator, :evaluation] }
+                           end
+
+    @submissions = @submissions.includes(association_preloads)
```
stepchud: there's an rspec failure and the sort by assignee views have 500 internal server error but the simple fix is to use preload instead of includes. 
also getting a bullet warning on the other views because of excluding the `evaluation_scores` from the query.
Screenshot 2025-02-25 at 5 02 12 PM
this works for both:
```suggestion
    association_preloads = { evaluator_submission_assignments: [:evaluator, { evaluation: :evaluation_scores }] }
    @submissions = @submissions.preload(association_preloads)
```

stepchud (commented):
very nice, approved by me but I can't approve my own PR. Feel free to approve and merge @emmabjj 👍

emmabjj (approved):
:shipit:

PR #438, opened 2025-02-21
Fix Mailer Hostname Mismatch

emmabjj:
Fix the mailer hostname mismatch in `config/initializers/mailer.rb`

Use the host set and default to challenge.gov instead of gsa.gov.


stepchud (approved):
we'll give it a shot

PR #445, opened 2025-02-26
Fix Mailer Template Accessibility & Link Color

emmabjj:
Fix mailer template accessibility (language and alt text) and fix the link color override by gmail.


stepchud (approved):
works for me, thanks for fixing the flaky spec!

Issue #388, opened 2025-02-05
Evaluation Panel Page Updates


r-bartlett-gsa:
### User Story
As a challenge manager, in order to accomplish all tasks associated with my challenge management in a consistent way, I would like evaluation panel workflows to be consistent in design and styles with other eval site features.

**Acceptance Criteria:**

- [x] When the challenge manager clicks on "Invite evaluators" or "[number] evaluators" links on the new challenges dashboard page, they are taken to the evaluation panel page associated with that challenge
- [x] Evaluation Panel page H1 is Challenge Title (if it's a phased challenge, the phase is indicated in the H1)
- [x] Page H2 is "Evaluation Panel"
- [x] Add Evaluators form consists of two fields:
  - [x] Full Name
  - [x] Email Address
  - [x] All fields are required
  - [x] Add Evaluator button adds the evaluator to the evaluator table with appropriate name, user role, and user status (no changes)
  - [x] Cancel button clears out the form
- [x] If there are no evaluators in the panel, a warning alert is displayed indicating so
- [x] Evaluation Panel table consist of the following headers:
  - [x] Name
    - [x] Links to the Evaluator's Submission Page
  - [x] Email Address
  - [x] User Role
  - [x] User Status
    - [x] Invite Sent status also includes an option to resend the invite
  - [x] # of Assigned Submissions
    - [x] If the number of assigned submissions is >0, then it links to the Evaluator's Submissions Page
  - [x] Action
    - [x] Delete button is displayed for each row
    - [x] Delete action shows confirmation modal
    - [x] Successful deletion shows a success alert on the elevation panel page
      - [x] Delete button removes an evaluator from the panel and all associations to submission evaluation (no changes)
- [x] Info alert is displayed for how to assign evaluators to submissions and a link to the submissions list page for that challenge
- [x] Designs (colors, fonts, labels, etc) match wireframes in **Figma:**  https://www.figma.com/design/sLkJM1Ua9Zu3HW2h4mVdgr/Challenge.gov-Eval-(Updated-01%2F27%2F25)?node-id=7003-153455&t=G2WnCXJa1OXhZSU3-1

### Definition of Done

**Doing (dev team)**
- [x] Code complete 
- [x] Code is organized appropriately
- [x] Any known trade offs are documented in the associated GH issue
- [x] Code is documented, modules, shared functions, etc.
- [x] Automated testing has been added or updated in response to changes in this PR
- [x] The feature is smoke tested to confirm it meets requirements
[...]

PR #481, opened 2025-03-12
[388] Fix Evaluator Panel Error Messages and Form Clearing

emmabjj:
Linked ticket: #388 

Address these AC failures/suggestions:
- [x] Update error messages in error alert and error message by the form field to: "Please provide evaluator's full name" and "Please provide evaluator's valid email address"
- [x] Should clicking on cancel clear out the form fields and also error messages (for example if the user changed their mind about adding a new evaluator, they should be able to clear out the form and not have prior error messages persist)?
- [x] Delete buttons should align in a column

Note: generateErrorMessage doesn't validate full_name as expected because first_name and last_name are the expected attributes on the model and not full_name. The inline_error message has to include last_name. 

***********

Screenshot 2025-03-11 at 7 13 46 PM

Screenshot 2025-03-11 at 7 22 18 PM

Screenshot 2025-03-11 at 7 22 34 PM

388_clear_errors_on_cancel

### DIFF
--- a/app/javascript/controllers/form_validation_controller.js
+++ b/app/javascript/controllers/form_validation_controller.js
@@ -67,4 +67,44 @@ export default class extends Controller {
       errorElement.innerHTML = message;
     }
   }
+  
+  clearForm(e) {
+    e.preventDefault();
+    const form = e.target.closest('form');
+
+    form.reset();
+    
+    form.querySelectorAll('input[type="text"], input[type="email"]').forEach(input => {
+      input.value = '';
+    });
+    
+    this.clearAllErrors(form);
+  }
+  
+  clearAllErrors(form) {
+    const errorAlert = form.querySelector('.usa-alert--error');
+    if (errorAlert) {
+      errorAlert.remove();
+    }
+  
+    const formGroups = form.querySelectorAll('.usa-form-group');
+    formGroups.forEach(group => {
+      const input = group.querySelector('input, textarea, select');
+      if (input) {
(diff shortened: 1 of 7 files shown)


app/javascript/controllers/form_validation_controller.js
```diff
@@ -67,4 +67,44 @@ export default class extends Controller {
...
+  
+  clearForm(e) {
+    e.preventDefault();
+    const form = e.target.closest('form');
+
+    form.reset();
+    
+    form.querySelectorAll('input[type="text"], input[type="email"]').forEach(input => {
+      input.value = '';
+    });
```
stepchud, line 79: isn't this accomplished by `form.reset()`?

stepchud: nevermind I tried it, guess not

app/models/evaluator_invitation.rb
```diff
@@ -18,9 +18,8 @@ class EvaluatorInvitation < ApplicationRecord
   belongs_to :challenge
   belongs_to :phase
 
-  validates :first_name, presence: true
-  validates :last_name, presence: true
-  validates :email, presence: true, format: { with: URI::MailTo::EMAIL_REGEXP }
+  validate :validate_full_name
+  validate :validate_email
```
stepchud, line 22: made these custom errors so they would appear correctly with translation

stepchud (approved):
:shipit:

PR #483, opened 2025-03-12
[400] | Adjust Reset Sort & Filter behavior

emmabjj:
Linked ticket: #400 

Fixes that address this feedback:

> When using Reset button, the focus should remain on the UI, instead of closing the UI and going to the top of the page.

When the user applies `reset` on a filtered submission list, the `reset` now:
1.) Keeps the sort & filter dropdown UI open 
2.) Keeps focus on _reset_ in the dropdown (for accessibility)
3.) Clears the filter/sort and search by submission id options from the UI in the dropdown menu
4.) Clears the filter/sort and search by submission id query params from the URL 
5.) Replaces the submissions table body with the reset (without filters/sorts/submission id) submissions data 

*******

reset_sort_and_filter

mobile_reset_sort_and_filter

### DIFF
--- a/app/javascript/controllers/sort_filter_menu_controller.js
+++ b/app/javascript/controllers/sort_filter_menu_controller.js
@@ -63,13 +63,36 @@ export default class extends Controller {
 
     window.location.href = `${window.location.pathname}?${queryParams}`
   }
-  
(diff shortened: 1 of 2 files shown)


stepchud (approved):
🤩

Issue #349, opened 2025-01-07
IAL2 verification for non-gov users


r-bartlett-gsa:
### User Story
As a security officer, in order to ensure the security of sensitive information and privacy of users on Challenge.gov, I would like Challenge.gov to verify user digital identities using facial matching comparison identity verification service provided by Login.gov and comply with NIST 800-63 IAL2 compliance.

**Acceptance Criteria:**

  - [x] Login.gov verification banner is displayed on the following pages:
    - [x] The dashboard page for challenge manager and evaluator roles
    - [x] #490 
    - [x] Submission evaluation page for evaluators
- [x] When the user clicks on Login.gov verification banner, the user is sent to login.gov to complete identity verification with facial matching comparison
- [x] If the user is successfully verified, the user gains access to submission and public solvers data (same as gov user)
- [x] If the user is not verified, the user is sent to failure to proof page, #346 
  - [x] The user continues to have IAL1 level access on challenge.gov

### Definition of Done

**Doing (dev team)**
- [x] Code complete 
- [x] Code is organized appropriately
- [x] Any known trade offs are documented in the associated GH issue
- [x] Code is documented, modules, shared functions, etc.
- [x] Automated testing has been added or updated in response to changes in this PR
- [x] The feature is smoke tested to confirm it meets requirements
- [x] Database changes have been peer reviewed for index changes and performance bottlenecks
- [x] PR that changes or adds UI
  - [x] include a screenshot of the WAVE report for the altered pages
  - [x] Confirm changes were validated for mobile responsiveness
- [x] PR approved / Peer reviewed 
- [x] Security scans passed
- [x] Automate accessibility tests passed
- [x] Build process and deployment is automated and repeatable
- [x] Feature toggles if appropriate
- [x] Deploy to staging
- [x] Move card to testing column in the board
---
**Staging**
- [ ] Accessibility tested (Marni)
  - [ ]  Keyboard navigation
  - [ ]  Focus confirmed
  - [ ] Color contrast compliance
  - [ ] Screen reader testing 
- [ ] Usability testing: mobile and desktop (Tracy or Marni)
- [ ] Cross browser testing - UI rendering is performant on below listed devices/browsers (Tracy or Marni)
  - [ ] Windows/Chrome
  - [ ] Windows/Edge
  - [ ] Mac/Chrome
  - [ ] Mac/Safari
  - [ ] iOS/Safari
- [ ] AC review (Renata)
[...]

Issue #346, opened 2025-01-07
Failure to Proof Page


r-bartlett-gsa:
### User Story
As a non gov user, in order to understand what information I have access to on Challenge.gov, I would like identity verification process explained/detailed with options available to me when my identity verification is unsuccessful.

**Acceptance Criteria:**

- [x] Users who fail IAL2 level identity verification with login.gov are sent to a page that displays failure message/explanation with contact information for login.gov and challenge.gov
- [x] Users who fail IAL2 can continue to authenticate at IAL 1 level

### Definition of Done

**Doing (dev team)**
- [x] Code complete 
- [x] Code is organized appropriately
- [x] Any known trade offs are documented in the associated GH issue
- [x] Code is documented, modules, shared functions, etc.
- [x] Automated testing has been added or updated in response to changes in this PR
- [x] The feature is smoke tested to confirm it meets requirements
- [x] Database changes have been peer reviewed for index changes and performance bottlenecks
- [x] PR that changes or adds UI
  - [x] include a screenshot of the WAVE report for the altered pages
  - [x] Confirm changes were validated for mobile responsiveness
- [x] PR approved / Peer reviewed 
- [x] Security scans passed
- [x] Automate accessibility tests passed
- [x] Build process and deployment is automated and repeatable
- [x] Feature toggles if appropriate
- [x] Deploy to staging
- [x] Move card to testing column in the board
---
**Staging**
- [x] Accessibility tested (Marni)
  - [x]  Keyboard navigation
  - [x]  Focus confirmed
  - [x] Color contrast compliance
  - [x] Screen reader testing 
- [x] Usability testing: mobile and desktop (Tracy or Marni)
- [x] Cross browser testing - UI rendering is performant on below listed devices/browsers (Tracy or Marni)
  - [x] Windows/Chrome
  - [x] Windows/Edge
  - [x] Mac/Chrome
  - [x] Mac/Safari
  - [x] iOS/Safari
- [ ] AC review (Renata)
- [ ] Deploy to production (production-like environment for eval capability) (dev team)
- [ ] Move to production column in the board
---
**Production**
- [ ] User and security documentation has been reviewed for necessary updates (Renata/Tracy/Dev team) 
- [ ] PO / PM approved (Jarah or Renata)
- [ ] AC is met and it works as expected (Jarah or Renata)
- [ ] Move to done column in the board (Jarah or Renata)

r-bartlett-gsa:
[...]

Issue #12, opened 2024-07-01
Login.gov IAL2 Integration


r-bartlett-gsa:
### User Story
As a security officer, in order to ensure the security of sensitive information and privacy of users on Challenge.gov, I would like Challenge.gov to verify user digital identities using facial matching comparison identity verification service provided by Login.gov and comply with NIST 800-63 IAL2 compliance.

**Acceptance Criteria:**

- [x] Challenge.gov identity authentication and verification with login.gov is configured as follows:
  - [x] Users with non .gov or .mil email accounts are defaulted as public solvers when they create an account
    - [x] Existing and new public solvers can login to challenge.gov at IAL1 level _**(current process)**_
  - [x] Users with .gov or .mil email address are defaulted as challenge managers when they create an account
    - [x] Challenge managers are manually reviewed / updated to evaluator accounts / verified or recertified / approved by GSA _**(current process)**_
      - [x] Users with .gov email addresses who are set to public solver roles (for example state or local gov user, whose account get updated manually to public solver by GSA) do not require IAL2 verification and login to the system at IAL1 level 
    - [x] Challenge managers and evaluators with .gov and .mil email addresses are considered agency-verified and exempt from facial matching comparison identity verification via login.gov
  - [x] Users with non .gov or .mil email accounts that have been upgraded to challenge manager or evaluator roles login to challenge.gov at IAL1 level
    - [x] Users with non .gov or .mil accounts at IAL1 level have limited access to challenge.gov _**(current process)**_
    - [x] Users with non .gov or .mil accounts have an option to verify at IAL2 level, #349 

### Definition of Done

**Doing (dev team)**
- [x] Code complete 
- [x] Code is organized appropriately
- [x] Any known trade offs are documented in the associated GH issue
- [x] Code is documented, modules, shared functions, etc.
- [x] Automated testing has been added or updated in response to changes in this PR
- [x] The feature is smoke tested to confirm it meets requirements
- [x] Database changes have been peer reviewed for index changes and performance bottlenecks
- [x] PR that changes or adds UI
  - [x] include a screenshot of the WAVE report for the altered pages
  - [x] Confirm changes were validated for mobile responsiveness
- [x] PR approved / Peer reviewed 
- [x] Security scans passed
- [x] Automate accessibility tests passed
- [x] Build process and deployment is automated and repeatable
- [x] Feature toggles if appropriate
- [x] Deploy to staging
- [x] Move card to testing column in the board
---
**Staging**
[...]

PR #488, opened 2025-03-24
[12] Login.gov IAL2 Integration

cpreisinger:
#12 
#349 
#346


app/views/sessions/failure_to_proof.erb
stepchud, line 1: doesn't this file need the .html.erb extension?

spec/spec_helper.rb
```diff
@@ -93,6 +93,7 @@ def mock_login_gov(user, code = "ABC123") # rubocop:disable Metrics/AbcSize
   allow(login_gov).to receive(:exchange_token_from_auth_result).with(code).and_return(
     [{ email: user.email, sub: user.token }]
   )
+  allow(login_gov).to receive(:authorization_url).with(ial2: true)
```
stepchud: do we always want to mock this as true? seems like most users, specifically .gov and .mil users, will not be receiving `ial2=true` response from login.gov. is that correct?
i'm thinking maybe this should be a parameter of the `mock_login_gov` function, or added as a different mock function e.g. `mock_login_gov_ial2`. let me know if you disagree?

config/routes.rb
```diff
@@ -3,6 +3,7 @@
 Rails.application.routes.draw do
   # Define your application routes per the DSL in https://guides.rubyonrails.org/routing.html
   get 'auth/result', to: 'sessions#result'
+  get 'auth/failure_to_proof', to: 'sessions#failure_to_proof'
```
stepchud, line 6: can we add a request spec for validating the contents of this route?

spec/system/evaluations_spec.rb
```diff
@@ -87,6 +94,21 @@
...
       end
+
+      it "shows the identity verification banner if not ial_level 2" do
+        visit phases_path
+
+        expect(page).to have_css(".usa-alert--info",
+                                 text: "To view submission information on Challenge.gov, you must verify your identity with Login.gov")
+      end
+
+      it "does not show the identity verification banner if ial_level 2" do
```
stepchud, line 105: 👍

app/views/sessions/failure_to_proof.erb
```diff
@@ -0,0 +1,16 @@
+<%= render 'shared/alert_error', alert_heading: t('alerts.failure_to_proof.heading'), alert_text: t('alerts.failure_to_proof.text') %>
+
+<div>
+  <h3>What you can do:</h3>
```
stepchud: shouldn't this be h1 since it's the only header on the page?

app/views/sessions/failure_to_proof.erb
```diff
@@ -0,0 +1,16 @@
...
+  <p>
+    For problems with your Login.gov account, please review <a href="https://login.gov/help/" target="_blank">Login.gov's Help Section</a>
+    or contact them at <a href="https://login.gov/contact/" target="_blank">login.gov/contact</a>.
+  </p>
+  <p>
+    If you've taken steps above and still can't verify your identity, please contact <a href="mailto:team@challenge.gov" target="_blank">team@challenge.gov</a>
```
stepchud: typo
```suggestion
    If you've taken steps above and still can't verify your identity, please contact <a href="mailto:team@challenge.gov" target="_blank">team@challenge.gov</a>.
```

app/views/sessions/failure_to_proof.erb
```diff
@@ -0,0 +1,16 @@
...
+  </p>
+  <p>
+    You still may have access to the Challenge.gov platform without verifying your identity, but your access will be limited to non-sensitive
+    information. Please proceed to login at the top of thte page.
```
stepchud: typo
```suggestion
    information. Please proceed to login at the top of the page.
```

stepchud (commented):
Initial code review, starting to test the feature now..

spec/spec_helper.rb
cpreisinger: Some of my specs were failing if I removed this line. Like that authorization_url method wasn't expecting to be able to receive ial2 as a param. I wasn't sure if this actually affected what was returned in the specs.

app/views/sessions/failure_to_proof.erb
cpreisinger: The `shared/alert_error` has an `h3` because of its use on other pages

cpreisinger: But the page needs an `h1` and `axe_clean` doesn't seem to complain about the out of order headings

app/views/sessions/failure_to_proof.html.erb
```diff
@@ -0,0 +1,16 @@
...
+  <p>
+    For problems with your Login.gov account, please review <a href="https://login.gov/help" target="_blank">Login.gov's Help Section</a>
+    or contact them at <a href="https://login.gov/contact" target="_blank">login.gov/contact</a>.
+  </p>
+  <p>
+    If you've taken steps above and still can't verify your identity, please contact <a href="mailto:team@challenge.gov" target="_blank">team@challenge.gov</a>.
+  </p>
+  <p>
+    You still may have access to the Challenge.gov platform without verifying your identity, but your access will be limited to non-sensitive
+    information. Please proceed to login at the top of the page.
```
stepchud: I think the message is a little misleading because they have already logged in here. Let's remove this last sentence.
```suggestion
    information.
```

stepchud (approved):
Tested failure_to_proof, rejected verification request and successful ial2 as a Challenge Manager. "Verify with Login.gov" banner is present in the locations specified.
Looks good to me.

PR #496, opened 2025-04-03
Fix intermittent submission_details_spec failure

stepchud:
1. Fixes the flaking `spec/system/submission_details_spec.rb`, which seems to have started when we switched the page to automatically save submission when the `eligible-for-evaluation` button is clicked.

1. Moves the downloaded csv files to a /tmp folder so it won't pollute the devs' `~/Downloads` folder (the Chrome default download location).

1. Fixes the system spec `skip_to_main_content_spec` that fails locally, intermittently.


emmabjj (approved):
Looks good! 👍

PR #504, opened 2025-04-07
[486] implement the site-wide banner

stepchud:
Adds the site-wide-banner to the Rails pages.

Banner is rendered when there is SiteContent that is present the current date is between the begin and end dates.

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -4,15 +4,15 @@
 #
 # Table name: submission_documents
 #
-#  id                 :bigint           not null, primary key
-#  user_id            :bigint           not null
-#  submission_id      :bigint
-#  filename           :string(255)      not null
-#  uuid               :uuid             not null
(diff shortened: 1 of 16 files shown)


app/models/document.rb
```diff
@@ -4,15 +4,15 @@
...
-#  uuid               :uuid             not null
-#  extension          :string(255)      not null
-#  name               :string(255)
-#  inserted_at        :datetime         not null
-#  updated_at         :datetime         not null
+#  id            :bigint           not null, primary key
+#  user_id       :bigint           not null
+#  submission_id :bigint
+#  filename      :string(255)      not null
+#  key           :uuid             not null
```
stepchud, line 11: these schema comment updates are auto-generated via `bundle exec annotate --models`

cpreisinger (approved):

Issue #489, opened 2025-03-24
Update global header/footer layout styles


stepchud:
# What/Why
As a product owner, I want the page layouts to match across Jekyll/Rails/Phoenix as much as possible.
This task specifically focuses on extending the backgrounds of the header/footer so that the Jekyll app and Rails apps layouts matchup. We should also ensure the Phoenix app is up to date in terms of these styles.

### Screenshot Examples
There are no Figma that demonstrate this spec and it was not included in the initial round of MMP development.

## Acceptance Criteria
- [x] Rails utility nav background extends to the width of the page (uncontained)
- [x] Rails footer is updated to use `widescreen` container to match the rest of the page content
- [x] Jekyll "An official website of the USG" banner matches Rails logged-in
  - [x] expand/collapse icon
  - [x] matching copy 
  - [x] matching layout, element spacing, font-weight etc
  - Rails:
  ![Image](https://github.com/user-attachments/assets/b3a92b87-959e-48ea-86df-f736b085584d)
  - Jekyll:
  ![Image](https://github.com/user-attachments/assets/402a1260-d957-4a67-87d8-cb9b9d69b045)

- [x] update Jekyll USWDS container size to `widescreen` to match the Rails setting
  - [x] default container width applies to `.usa-nav-container` and `.grid-container`
  - ![Image](https://github.com/user-attachments/assets/5afa749a-4e70-4706-b3eb-5783e9a2b880)
  - ![Image](https://github.com/user-attachments/assets/f57bbbe4-5bc6-422b-99fe-475b69f0df14)

_Disregard the following if this work **does not** include any UI changes_
- [x] WAVE screenshot
- [x] Automated a11y tests (axe-rspec specs)
- [x] Keyboard accessibility (tab navigation, form interactions)
- [x] Responsiveness at desktop and mobile

stepchud:
WAVE:
![Image](https://github.com/user-attachments/assets/2a13b36d-f472-4554-821f-b8191f42135d)

afeijoo:
Archiving repo; no longer applicable.

PR #512, opened 2025-04-10
[489] Update global header/footer layout styles

cpreisinger:
#489

### DIFF
--- a/app/views/shared/_flash.html.erb
+++ b/app/views/shared/_flash.html.erb
@@ -7,7 +7,7 @@
       else "usa-alert--info"
       end %>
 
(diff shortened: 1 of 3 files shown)


app/views/shared/_flash.html.erb
```diff
@@ -7,7 +7,7 @@
       else "usa-alert--info"
       end %>
 
-    <div class="grid-container usa-alert <%= alert_class %> usa-alert--slim">
+    <div class="usa-alert <%= alert_class %> usa-alert--slim">
```
stepchud:
```suggestion
    <div class="grid-container usa-alert <%= alert_class %> usa-alert--slim">
```

stepchud (approved):

Issue #511, opened 2025-04-10
User activity on Phoenix app resets the Rails session timeout


stepchud:
# What/Why
When the user interacts with the Phoenix app, it should refresh and keep-alive the Rails session.
See #507 for a reference implementation on Rails

NOTE: This functionality is not critical for Evaluation since most Eval activity takes place on the Rails app. Therefor this work was deprioritized for the Eval module as it related to updating Phoenix.

### Figma Link
No design.

## Acceptance Criteria
- [ ] When the user interacts with Phoenix site, a request is made to Rails
- [ ] Rails updates the user's cookie when the request is processed

_Disregard the following if this work **does not** include any UI changes_
- [ ] WAVE screenshot
- [ ] Automated a11y tests (axe-rspec specs)
- [ ] Keyboard accessibility (tab navigation, form interactions)
- [ ] Responsiveness at desktop and mobile

afeijoo:
Archiving repo; no longer necessary.

PR #514, opened 2025-04-11
[511] User activity on Phoenix app resets the Rails session timeout

cpreisinger:
#511


afeijoo:
Archiving repo; no longer necessary.

